PR #379, opened 2016-04-20
Fix fullscreen - remove all the garbage which breaks things and let G…

jkufner:
Remove all the garbage which breaks things and let Gtk do its job.

The most problematic thing was the `gdk_window_set_override_redirect` call. That was what made most of the mess. Another bad idea was to deny resizing of the fullscreen window, which causes window manager to fail to adjust window size to screen size.

Tested on KDE5. I really doubt it can be worse than it is now, but feel free to test it anyway :)


jkufner:
Today I tested it on multi-screen setup, works fine.

PR #415, opened 2016-07-09
Fix hidden directories not showing in a tree view

maciejsszmigiero:
Directory tree view contains code that when "Show hidden files or folders"
option is not enabled but the user enters a path containing hidden
directories the missing tree entries are added automatically.

This didn't work (at least) in case when the first hidden directory in
newly selected path was in a current directory because the tree populating
function exited early when directory contents did not change since last
run, before checking for this special case.

Fix this by skipping this check when we have to add such hidden directory
entry.

This check must also be skipped when 'force' flag is passed to this
function, otherwise when enabling "Show hidden files or folders" option
the missing directories are not added to the tree until Geeqie is
restarted.


maciejsszmigiero:
Ping.

This is a simple fix.

PR #423, opened 2016-07-28
Fix #414 Render Intents

### LINKED ISSUE #414: Wrong rendering intent when using monitor profile
*Copied from http://www.mail-archive.com/geeqie-devel@lists.sourceforge.net/msg01816.html*

Hello

There is quite a serious bug or lack of feature in Geeqie. One can use a 
monitor profile, but the monitor profile uses the perceptual rendering intent 
instead of relative colorimetric, if it supports both. The default option 
should always be relative colorimetric, even if more rendering intents are 
present in the profile. That is of utmost importance and more of a bug. 
Secondary to that is adding an easily-accessible option to switch rendering 
intents, more of a feature request.

This is my monitor profile which when used in Geeqie leads to very wrong 
colors because the perceptual intent is used. When I use it in RawTherapee, 
GIMP or XnViewMP I can select relative colorimetric, and then things are 
good.
https://filebin.net/jh38xbpadwa0xmku/latest.icc

I hope this issue gets addressed soon, because until such a time I must 
resort to other image viewers when I need color management.

Kind regards

cclark replied:
> At around line 187 of color-man.c is:
```
    cc->transform = cmsCreateTransform(cc->profile_in,
                       (has_alpha) ? TYPE_RGBA_8 : TYPE_RGB_8,
                       cc->profile_out,
                       (has_alpha) ? TYPE_RGBA_8 : TYPE_RGB_8,
                       INTENT_PERCEPTUAL, 0);
```
`INTENT_PERCEPTUAL` is what lcms2 uses as default in its documentation. 
If it were simply a matter of swapping that to `INTENT_RELATIVE_COLORIMETRIC`, it would be easy to put an option in the preferences tab.
In fact, if lcms2 has a "`cmsIsIntentSupported`" function, it would be possible to put a drop-down box of the available intents.

caclark:
https://github.com/BestImageViewer/geeqie/issues/414

Permit the user to select the rendering intent.

### DIFF
--- a/src/color-man.c
+++ b/src/color-man.c
@@ -188,7 +188,7 @@ static ColorManCache *color_man_cache_new(ColorManProfileType in_type, const gch
 					   (has_alpha) ? TYPE_RGBA_8 : TYPE_RGB_8,
 					   cc->profile_out,
 					   (has_alpha) ? TYPE_RGBA_8 : TYPE_RGB_8,
-					   INTENT_PERCEPTUAL, 0);
+					   options->color_profile.render_intent, 0);
 
 	if (!cc->transform)
 		{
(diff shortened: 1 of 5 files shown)


Beep6581:
Please merge this before making a new release.

Beep6581:
Please merge this.

PR #428, opened 2016-08-19
Fix memory leak

xi2:
Hi.

I think I found another code path, similar to that fixed in 03283a3, where `cairo_create` may get called and there is a possible return from the function (`rt_border_draw` in this case) without a corresponding `cairo_destroy`. The pull request here addresses this.

Thoughts?

Thank you for maintaining geeqie.

### DIFF
--- a/src/cellrenderericon.c
+++ b/src/cellrenderericon.c
@@ -692,7 +692,9 @@ gqv_cell_renderer_icon_render(GtkCellRenderer		*cell,
 
 	if (!pixbuf && !text)
 		{
+#if !GTK_CHECK_VERSION(3,0,0)
 		cairo_destroy(cr);
+#endif
 		return;
 		}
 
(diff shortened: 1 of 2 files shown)


xi2:
I write "Thoughts?" with a question mark because I am not really familiar with cairo programming. I notice that [here](https://github.com/BestImageViewer/geeqie/blob/03283a34699c8b88f08b39dee90ae13ee1d63e58/src/cellrenderericon.c#L875) `cairo_destroy` is only called for `!GTK_CHECK_VERSION(3,0,0)`, and I don't know why it would not always be called for all GTK versions, so I am not 100% confident of what is going on with create/destroy.

xi2:
Sorry to reply to myself. I see now why the `cairo_destroy` mentions previously is wrapped in the GTK version check CPP macro. It is because the `cairo_create` is. This would mean that the fix in 03283a3 is probably incorrect for GTK3. I have added another commit to this PR to address that.

PR #452, opened 2016-12-25
Fix compile warnings, and add -Werror to the default CFLAGS

xsdg:
More specifically, sets default CFLAGS to "-g -O2 -Werror -Wno-error=deprecated-declarations"

### DIFF
--- a/configure.in
+++ b/configure.in
@@ -17,6 +17,9 @@ dnl GNU General Public License for more details.
 AC_PREREQ(2.57)
 AC_INIT([geeqie], m4_esyscmd_s(git rev-parse --quiet --verify --short master), [geeqie-devel@lists.sourceforge.net], [], [http://www.geeqie.org/])
 
+# Add -Werror to the default CFLAGS
+CFLAGS+=" -Werror -Wno-error=deprecated-declarations"
+
 # Check for rightly dirs
 AC_CONFIG_SRCDIR([src/main.c])
 
--- a/src/cache_maint.c
+++ b/src/cache_maint.c
@@ -32,6 +32,7 @@
 #include "ui_spinner.h"
 #include "ui_tabcomp.h"
 #include "ui_utildlg.h"
+#include "window.h"
 
 
 typedef struct _CMData CMData;
--- a/src/layout.c
+++ b/src/layout.c
@@ -1728,7 +1728,7 @@ gboolean layout_tools_float_get(LayoutWindow *lw, gboolean *popped, gboolean *hi
 void layout_toolbar_toggle(LayoutWindow *lw)
 {
 	if (!layout_valid(&lw)) return;
-	if (!lw->toolbar) return;
+	if (!lw->toolbar[TOOLBAR_MAIN]) return;
 
 	lw->options.toolbar_hidden = !lw->options.toolbar_hidden;
 
@@ -2125,8 +2125,8 @@ LayoutWindow *layout_new_with_geometry(FileData *dir_fd, LayoutOptions *lop,
 
 	layout_config_parse(lw->options.style, lw->options.order,
 			    &lw->dir_location,  &lw->file_location, &lw->image_location);
-	if (lw->options.dir_view_type >= VIEW_DIR_TYPES_COUNT) lw->options.dir_view_type = 0;
-	if (lw->options.file_view_type >= VIEW_FILE_TYPES_COUNT) lw->options.file_view_type = 0;
+	if (lw->options.dir_view_type > DIRVIEW_LAST) lw->options.dir_view_type = 0;
+	if (lw->options.file_view_type > FILEVIEW_LAST) lw->options.file_view_type = 0;
 
 	/* divider positions */
 
--- a/src/layout_util.c
+++ b/src/layout_util.c
@@ -2134,7 +2134,7 @@ void layout_actions_setup(LayoutWindow *lw)
 					   menu_split_radio_entries, G_N_ELEMENTS(menu_split_radio_entries),
 					   0, G_CALLBACK(layout_menu_split_cb), lw);
 	gtk_action_group_add_radio_actions(lw->action_group,
-					   menu_view_dir_radio_entries, VIEW_DIR_TYPES_COUNT,
+					   menu_view_dir_radio_entries, DIRVIEW_LAST + 1 /* count */,
 					   0, G_CALLBACK(layout_menu_view_dir_as_cb), lw);
 	gtk_action_group_add_radio_actions(lw->action_group,
 					   menu_color_radio_entries, COLOR_PROFILE_FILE + COLOR_PROFILE_INPUTS,
--- a/src/preferences.c
+++ b/src/preferences.c
@@ -410,7 +410,7 @@ static void config_window_close_cb(GtkWidget *widget, gpointer data)
 
 static void config_window_help_cb(GtkWidget *widget, gpointer data)
 {
-	GtkWidget *notebook = GTK_NOTEBOOK(data);
+	GtkWidget *notebook = GTK_WIDGET(data);
 	gint i;
 
 	static gchar *html_section[] =
@@ -426,7 +426,7 @@ static void config_window_help_cb(GtkWidget *widget, gpointer data)
 	"GuideOptionsBehavior.html"
 	};
 
-	i = gtk_notebook_get_current_page(notebook);
+	i = gtk_notebook_get_current_page(GTK_NOTEBOOK(notebook));
 	help_window_show(html_section[i]);
 }
 
@@ -1979,7 +1979,7 @@ static void intent_menu_cb(GtkWidget *combo, gpointer data)
 }
 
 static void add_intent_menu(GtkWidget *table, gint column, gint row, const gchar *text,
-			     guint option, guint *option_c)
+			     gint option, gint *option_c)
 {
 	GtkWidget *combo;
 	gint current = 0;
--- a/src/typedefs.h
+++ b/src/typedefs.h
@@ -38,12 +38,18 @@ typedef enum {
 
 typedef enum {
 	DIRVIEW_LIST,
-	DIRVIEW_TREE
+	DIRVIEW_TREE,
+
+	// Keep this up to date!
+	DIRVIEW_LAST = DIRVIEW_TREE
 } DirViewType;
 
 typedef enum {
 	FILEVIEW_LIST,
-	FILEVIEW_ICON
+	FILEVIEW_ICON,
+
+	// Keep this up to date!
+	FILEVIEW_LAST = FILEVIEW_ICON
 } FileViewType;
 
 #define	CMD_COPY     "geeqie-copy-command.desktop"
--- a/src/view_dir.h
+++ b/src/view_dir.h
@@ -30,8 +30,6 @@ enum {
 	DIR_COLUMN_COUNT
 };
 
-#define VIEW_DIR_TYPES_COUNT 2
-
 ViewDir *vd_new(DirViewType type, FileData *dir_fd);
 
 void vd_set_select_func(ViewDir *vdl, void (*func)(ViewDir *vdl, FileData *fd, gpointer data), gpointer data);
--- a/src/view_file.h
+++ b/src/view_file.h
@@ -21,8 +21,6 @@
 #ifndef VIEW_FILE_H
 #define VIEW_FILE_H
 
-#define VIEW_FILE_TYPES_COUNT 2
-
 #define VFLIST(_vf_) ((ViewFileInfoList *)(_vf_->info))
 #define VFICON(_vf_) ((ViewFileInfoIcon *)(_vf_->info))
 


gdt:
Have you verified that this works with clang as well?  While I'm pretty sure that most of these flags are not specified by C99, it seems there is a de facto subset for warnings that compilers other than gcc implement.

It would be nice to split into two commits, one for fixes and  one for changing the flags.  It seems reasonably likely that some people will have to revert the flag change at least temporarily.

xsdg:
gdt: Good point; done.  (And no, I haven't checked with clang)

We'll see how things go, but if the full -Werror ends up being untenable, I think we'd catch a bunch of serious issues just with:
-Werror=implicit-function-declaration -Werror=incompatible-pointer-types

gdt:
If you haven't tested with clang, I think it would be good to wait until someone has, or to add an if statement around to only add the not-standardized error flags to gcc, for now.

And separately, if you did a build with clang, you might find some other warnings that you aren't seeing now.  It seems that a lot of things are caught by both, but that switching in either direction usually finds new things.

xsdg:
Thanks for poking me about due diligence.  Laziness is fun but not always the best option.

Turns out that Clang has a bug where they assume that enum values cannot be out-of-range (which is true for C++, but not true for C), which causes it to emit a spurious warning.  Because it's a warning that GCC doesn't understand, clang users will need to uncomment a line in configure.in in order to compile.

If someone with more autoconf-fu can make this happen automatically, that would be awesome.  I'm content with this solution, though, particularly since Clang for C seems to be generally pretty uncommon.

gdt:
clang is the default compiler on Mac and I believe also on FreeBSD.   In pkgsrc, people  do bulk builds with clang on NetBSD, too.

I wonder if it would be better to clean up the enum stuff; I am not sure exactly what out-of-range enum means, but if it's an enum variable holding a value which is not one of the declared enum values, that seems more or less wrong even if legal.    There's a long tradition of cleaning up legal code because gcc doesn't like it!

autoconf is really just shell and macros.  So if you have a variable with the compiler name already, or a variable that says we have gcc, then you can just

if [ "$isgcc" = "yes" ]; then
   whatever you had to add the troublesome warnings
fi

and then the extra warnings which are only known to be ok with gcc will only get added for gcc.

mowgli:
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA512

Hi Folks,

Am Di den 27. Dez 2016 um 14:09 schrieb Greg Troxel:
> I wonder if it would be better to clean up the enum stuff; I am not sure exactly what out-of-range enum means, but if it's an enum variable holding a value which is not one of the declared enum values, that seems more or less wrong even if legal.    There's a long tradition of cleaning up legal code because gcc doesn't like it!

Independent, where that compiler is standard, it should work with it as
it is very common.

On the other hand, (I just followed this with one eye) I also think we
should fix that out of bound use. Do you have some pointers where it
happens?

> autoconf is really just shell and macros.

And some magic. ;-)

Regards
   Klaus
- -- 
Klaus Ethgen                                       http://www.ethgen.ch/
pub  4096R/4E20AF1C 2011-05-16            Klaus Ethgen <Klaus@Ethgen.ch>
Fingerprint: 85D4 CA42 952C 949B 1753  62B3 79D0 B06F 4E20 AF1C
-----BEGIN PGP SIGNATURE-----
Comment: Charset: ISO-8859-1

iQGzBAEBCgAdFiEEMWF28vh4/UMJJLQEpnwKsYAZ9qwFAlhiwq8ACgkQpnwKsYAZ
9qx4fQv8C3MfHRGDIrjtHB2dl1S/xZQV2k/2mFTG9U8juSUr+7FM/dzYxe2Ab59k
uKhTznW42UAIetNyAS2QCbfOjwG2v9ILE12jyPkiGbK2kn8ZzR6tvID+edTWIfV8
WveqfqAKw8agIUGpB5hK+m9mt3ZZxkCEySl9dgQj+6fe5nwWo6hnM7a2XNkPsAiV
E8C8ZaI6Lv8WrMtFuFI2Ne/eHq6DdEzIFl+8BAJtyTtOg4N4qW7FzmACDFRXorgu
2lzDGLBt/UcHZM951XFzrKIn3aUOOQ55FQ1FPYk7TTlimo+tHjb74+vJbCe5MNye
/xyGAXPzvXMb4xqQkp6JsHPyqpV3Fxet8tr5RtmfAzOqaZ/Zr4rhRXU/LmWoBpKW
eF0UXdjlW/13pwNz9O3u7tc57p7NI1JQKW1VjCD9iTFzKtzNvc0GQVeNPNKPtKEe
ZVOzJBebw1FY+VEz245N/FJZVIdnKUkNeHenA/rhRgSA10NHg1wxvOJkVKn5lWM0
/KQGcVQU
=SjHw
-----END PGP SIGNATURE-----

xsdg:
Fundamentally, the check is valid.  That's why this is a Clang bug.  Fundamentally, it is legal in C for something, somewhere, to have set an enum to an invalid value.  And, consequently, it is legitimate for this code to double-check that the enum has a value that actually makes sense.

Here's the warning:

    layout.c:2128:32: error: comparison of constant 2 with expression of type 'DirViewType' is always false [-Werror,-Wtautological-constant-out-of-range-compare]
            if (lw->options.dir_view_type >= VIEW_DIR_TYPES_COUNT) lw->options.dir_view_type = 0;
                ~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~
    layout.c:2129:33: error: comparison of constant 2 with expression of type 'FileViewType' is always false [-Werror,-Wtautological-constant-out-of-range-compare]
            if (lw->options.file_view_type >= VIEW_FILE_TYPES_COUNT) lw->options.file_view_type = 0;
                ~~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~

You'll note that this is validation of a value that was read from a config file.  The Clang warning is essentially, "this check is unnecessary."  That is obviously fundamentally incorrect, given C's lack of guarantees about enum values.  You can't actually prevent someone from sticking invalid data into the rc file.  But the takeaway from this bug would be that you *also* can't validate the enum value in your source code, which is preposterous.

Consequently, the answer here is to ignore that warning in Clang and to leave the code as-is.

As for autoconf, the solution should be to check if the compiler is Clang, not to check if the compiler *isn't* gcc.

mowgli:
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA512

Am Di den 27. Dez 2016 um 22:21 schrieb Omari Stephens:
> Fundamentally, the check is valid.  That's why this is a Clang bug.  Fundamentally, it is legal in C for something, somewhere, to have set an enum to an invalid value.  And, consequently, it is legitimate for this code to double-check that the enum has a value that actually makes sense.
> 
> Here's the warning:
> 
>     layout.c:2128:32: error: comparison of constant 2 with expression of type 'DirViewType' is always false [-Werror,-Wtautological-constant-out-of-range-compare]
>             if (lw->options.dir_view_type >= VIEW_DIR_TYPES_COUNT) lw->options.dir_view_type = 0;
>                 ~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~
>     layout.c:2129:33: error: comparison of constant 2 with expression of type 'FileViewType' is always false [-Werror,-Wtautological-constant-out-of-range-compare]
>             if (lw->options.file_view_type >= VIEW_FILE_TYPES_COUNT) lw->options.file_view_type = 0;
>                 ~~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~

Ah, now I see it.

Ok, then it is clear that there is a clang bug.

However, There would be a simple but a bit dirty solution. Just having a
third enum value in DirViewType and FileViewType that is never used.
Maybe just for this compiler. :-(

Gruß
   Klaus
- -- 
Klaus Ethgen                                       http://www.ethgen.ch/
pub  4096R/4E20AF1C 2011-05-16            Klaus Ethgen <Klaus@Ethgen.ch>
Fingerprint: 85D4 CA42 952C 949B 1753  62B3 79D0 B06F 4E20 AF1C
-----BEGIN PGP SIGNATURE-----
Comment: Charset: ISO-8859-1

iQGzBAEBCgAdFiEEMWF28vh4/UMJJLQEpnwKsYAZ9qwFAlhi3VAACgkQpnwKsYAZ
9qxS5Av/VIENdl4BcSxNvBlgmofWS3eH8w15zrVZbHnJ1tsu7C/NEDSJrUz/2cEW
+E7wjJrkjLQI6seabHsz+w18ftzIqH5WVb2MNR9fywulq60ZY7G3e3Sjf6ygJmNP
tXfLEexCKdAeEf/g94GAVgDK4LBCU6dBiGu6sQeNHswtE7qkY1VsseLaKCs1S1bJ
fQkMrIlSYs74ctm03Nzn3ussncTmT8lhpuG6aws4vruu+E4oDk05Y+hroi1aU+dv
lEeQ1nT7hH+th8joJPgSBSFTYaVzxXRRHZvTiFbAm1tFjbczJLyCJqznksz3JHbW
K8QWS8j0hDQhMFjxnrL3EX2FQCprVXZibT8vh+0Axo6n19ZMq/b56DuldF97Twnj
jjM+2G1/CxF/xkGUNjXgyM+WPkUW7wxMr59S7D3FEC+PFThp5AHKgUAGQSwQfnWf
JbIQ6LIUepXotU6DWZcfP/wmgG6YeTF6+jnqQSWYS5GU2psIKnnQ9FHgMo+LX7l2
DlgPKOap
=XfmW
-----END PGP SIGNATURE-----

caclark:
On 27/12/16 21:21, Omari Stephens wrote:
> ||
>
> You'll note that this is validation of a value that was read from a 
> config file. The Clang warning is essentially, "this check is 
> unnecessary." That is obviously fundamentally incorrect, given C's 
> lack of guarantees about enum values. You can't actually prevent 
> someone from sticking invalid data into the rc file. But the takeaway 
> from this bug would be that you /also/ can't validate the enum value 
> in your source code, which is preposterous.
>
>
This is not a comment on any solution proposed here - I just point out 
that all external values should be checked and limited before being 
stored and used internally - standard practice on the real-time stuff I 
used to do.

(And in this case it would help readability if the count was defined 
next to the item it was counting)

I mention this only because I recently had a problem I think caused by 
switching between the various versions of Geeqie - all of which use the 
same configuration file. I somehow ended up with a faulty configuration 
value that affected all versions. This isn't a significant problem for 
users, and I didn't bother following it.

xsdg:
mowgli: I had considered that.  The downside there is that compilers _do_ enforce that switch statements over enums cover all enum values.  So we'd have to add a "case VIEW_DIR_TYPES_COUNT: break;" to every switch statement involving one of these types.

Actually, one trick that might work is setting the count enum value to equal the last item in the enum.  Then the switch statement would cover all possible values, and the compiler might be happy.  I'll give that a shot.

caclark: I agree.  And generally, the rcfile code needs to be rewritten.  It's one of the files that would benefit most from C++ features like templating and method overloading.

xsdg:
@mowgli, @gdt: I think this is ready to go now.  Compiles properly with clang-3.8 and clang-3.9 with no file modifications (as well as gcc, obviously).

gdt:
It looks like building with clang found some more actual issues.
This seems to add text about having to uncomment things and then remove it.   I don't know about geeqie norms, but in my projects we've done a rebase to clean that up.

So it looks like you'll end up with 1 commit each to fix actual problems found by gcc and then clang, 1 command to make enum stuff easier to follow and avoid the clang bug, and one to flip on warnings.

xsdg:
If it's just me, I'd totally rebase.  Still thinking of doing that.  GitHub appears to counter-recommend rebases for branches that are already pushed (possibly because it could mess up other folks who have pulled that branch), which is the main reason I didn't collapse things in any way.  I agree with you that I'd rather not have the intermediate states in the codebase, and I'll probably go ahead and rebase + force-push so that all of the eventually-committed commits are useful.

If other folks have thoughts on this, I'm all ears.

mowgli:
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA512

Hi Omari,

Am Di den  3. Jan 2017 um 23:58 schrieb Omari Stephens:
> If it's just me, I'd totally rebase.  Still thinking of doing that.  GitHub appears to counter-recommend rebases for branches that are already pushed (possibly because it could mess up other folks who have pulled that branch), which is the main reason I didn't collapse things in any way.  I agree with you that I'd rather not have the intermediate states in the codebase, and I'll probably go ahead and rebase + force-push so that all of the eventually-committed commits are useful.

It is a really bad idea to rebase stuff that is already shared with
others. You most likely will ask for troubles this way.

However, it is fully ok to rebase stuff you prepare for push/share and
that is only existing in your own sandbox.

So, please, never ever even think of rebasing stuff that is already
published.

Regards
   Klaus
- -- 
Klaus Ethgen                                       http://www.ethgen.ch/
pub  4096R/4E20AF1C 2011-05-16            Klaus Ethgen <Klaus@Ethgen.ch>
Fingerprint: 85D4 CA42 952C 949B 1753  62B3 79D0 B06F 4E20 AF1C
-----BEGIN PGP SIGNATURE-----
Comment: Charset: ISO-8859-1

iQGzBAEBCgAdFiEEMWF28vh4/UMJJLQEpnwKsYAZ9qwFAlhsSx0ACgkQpnwKsYAZ
9qwNfQv/bJ4ry9G++Aqzwo2obZnCKOKwg7BHRxPsJRLNjhD8h5EHkJCnO1ff1ACo
RpkI3gbEC1YAHXATsn8RwauUeCKGQ7HR1w6lqTsbU0eILrgUaMcWgHyLnF7OmnKg
vwedxpKZCEnj8k5C7Llry5qdHsWzunELt00/6BHMwd+kGL5tR7yxaGqK+RtEzcfX
rMONdy7ZVesj8pbkSBoX+tSp0DcPrvv3ujw885mBlLjf/ZT1hbHhj+ick8OuHv42
o2icDxxzAFqDXfPYDokuNNc8johZA/MUP8ewdi1TMuYwA+XQLraj9WZVT1UYbzlD
/SHvZyrDxGujLluQQhhEbJ+ORVZsxmXwvRCjgALVxHkJ0YjYMNGgNSFHvpENrHJV
MXChRyciFZFY/K8G2bK1FJkxfwJN6cmLzShOYgD250Fnd+7dKiEmtTRdgl7frDo8
RLokVOodEiQgsyDbpGiVVGdJLxChyzhwhXyqQHMRHpj6BlPWBQBoX3N7ii3R8ORD
Mt6FZilu
=zPyM
-----END PGP SIGNATURE-----

xsdg:
Okay, I think we're all on the same page, then.  I will rebase my xsdg/geeqie pull-request branch, since that's basically still in my playground.  But if this had already been pushed to the BestImageViewer/geeqie repository, then we'd just have to live with it.

xsdg:
Done.

gdt:
Glad it's rebased.  My take is that a branch that is proposed for merging can be rebased a lot, since no one should be deriving anything from it, and the review/tweak process is about making it better and cleaning it up.   Which repo it's in doesn't really matter to me; it's about what we have decided to be part of the history vs a draft that is intended to, when ready, be added to that official history.

FWIW, I ran a project where branches were rebased about 10 times each on the way to merging, but once merged to master we considered it immutable.  One thing we did was to put the rebased version in .rb1, .rb2, etc. to avoid confusion.   I don't know if pull requests can be switched over to the .rb1 version though.

All in all, I think this discussion/rebase process has been great and I suspect the mechanics will be a background activity in the future, now that two of you who are useful and me the lurker have discussed.

xsdg:
@mowgli Just so we're on the same page, should I feel free to merge this after folks say that it looks good?  Should I wait for you or someone else to hit the merge button?

mowgli:
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA512

Hi Omari,

Am So den  8. Jan 2017 um  2:20 schrieb Omari Stephens:
> @mowgli Just so we're on the same page, should I feel free to merge this after folks say that it looks good?  Should I wait for you or someone else to hit the merge button?

Go ahead. You are one of the devs like the others. ;-)

I just would like you to send me your ssh key so I could add your access
to the main repo too.

Git is pretty good for that kind of development. You can always go back
if something goes really wrong.

Regards
   Klaus
- -- 
Klaus Ethgen                                       http://www.ethgen.ch/
pub  4096R/4E20AF1C 2011-05-16            Klaus Ethgen <Klaus@Ethgen.ch>
Fingerprint: 85D4 CA42 952C 949B 1753  62B3 79D0 B06F 4E20 AF1C
-----BEGIN PGP SIGNATURE-----
Comment: Charset: ISO-8859-1

iQGzBAEBCgAdFiEEMWF28vh4/UMJJLQEpnwKsYAZ9qwFAlhxl2kACgkQpnwKsYAZ
9qwLTgv/WI0OcqXWLMPCi4br4TJnMaSJdOphj2ZAy04gtfBqJHi91HLroMTfp3Xy
YqteLsm59k17KgNPJtzEU+8ULaef9Dd33OHuyWRh86RyY2zG2gnBobMNG7KamIBs
WtnJK6VbgYyMZTEc3Wozv5RStSxlq+RQPMXZK6EHShUi9cqk9dQ1DncS4MhPr8Uf
bsG67+twxeuZ99XOAt4SKuDRPwF/COjIifSso0GDh6gUmaAU2ZsFCQrQsfFuPruT
Wqo8uW71l7mepaARMagO2QLG5lXKuBudrf6w+6o65bEQkymz5CDhlOB63GV6hExJ
ZrUxV8+U4G2V9mRbNtkyU0OfmH5zPP1xqITvS0hHZeCe0AU8Do9tO0TTfxkkQn0v
xoSRw30dIwczMVDtAXidKYnmQU3xzejeU3yWaQAEqPE1r+Y1BNM759pVF4nVUc38
/Y3zxOV8qNS3/UXFjjPY/Z/yUOXDn8fC+ywpZQJumLS75AIcWFlRIRFdkTS7vAz7
t3U3oCXZ
=+fMZ
-----END PGP SIGNATURE-----

PR #458, opened 2017-01-01
Revamp pan view filtering to support different modes and grouping.

xsdg:
Conceptually, the filter is a sequence of filter elements, which are considered
in order for each image in the pan view.  The end result is that each image is
either rejected or displayed as normal.

I'm thinking of adding a "Simple" versus "Advanced" mode.  "Simple" would disable Group mode, and would also disable regular expressions.  Advanced would be what we've got here.

It would also be nice to refactor this in an MVC style, where the filtering happens at/after the model level, and pan-view and the standard directory-view are just Views, in the technical sense.  nadvornik had this idea back in 2008, and the continued divergence between directory view and pan view further illustrates that it's a better option than what exists right now.

Depends on Pull Request #454


caclark:
Hello Omari
It's clear from all the work you are doing here that you are a heavy-duty user of Pan View.
I don't use Pan View, but if I did it would be to use the calendar view to see images I took on particular days. This would be far quicker and more visually intuitive than making multiple calls to the search window.

The reason I don't use it is:
a) I can't allocate keywords to a selection
b) I can't allocate a selection to a collection
c) if I use "Show Image", at say 10%, I can't use the mouse to slide the image around so that I can see the underlying thumbnails
d) I can't double-click on a thumbnail to bring up the image in the main window

Do you also experience these shortcomings, or do you use Pan View in a very different way?

xsdg:
(Moving discussion about @caclark's question to #461)

PR #473, opened 2017-02-19
A small parameter type fix for "motion-notify-event" signal handlers

maciejsszmigiero:
The "motion-notify-event" signal "event" parameter is a GdkEventMotion structure,
not GdkEventButton.
These are mostly compatible, let's use a correct one however.


caclark:
Implemented in commit 08f1031b6e

PR #493, opened 2017-06-25
fix implicit declaration in non debug build

casta:
The new debug functions get/set regex are only declared for debug builds.
This fixes the non debug one.


casta:
merged, thanks

PR #495, opened 2017-07-07
Merge IconData into FileData

xsdg:
This is the first step in creating a model/view separation in the view_file_* infrastructure.  Getting rid of IconData means that we can unify codepaths that previously differed between view_file_icon* and view_file_list*, which should go a long way toward helping us share the same model between view_file_icon, view_file_list, and pan-view.

At some point, it'll be useful to pull the view-specific bits out of FileData again, but it'll be easier to figure out exactly what that should look like once there exists a shared model.


caclark:
Any effort that goes into simplifying or modularising Geeqie is certainly worth doing.
Unfortunately I don't think I am competent to make any technical comment on these changes, but I do hope  that I may be able to take advantage of them in future.
Recently I tried to implement some multi-threading, but immediately got lost in the tangled jungle of viewfileicon, viewfilelist, viewfiledir, viewfileanything. I think Geeqie has a problem that I guess most people have seen - a project just grew-and-grew-and-grew until it was so complex that it couldn't be maintained (well, to me Geeqie is complex). Perhaps some high-level documentation would help, so I might try that. One day...

xsdg:
@caclark: Yeah.  Although I've been surprised at how much of the code I've been diving into that originated pre-2005.

I think multi-threading is one of a variety of things that would be way easier if our codebase weren't in C.  For instance, filelist_sort_full uses static global variables for sort params, specifically because the g_list_sort function doesn't allow you to pass in extra info — it takes a list and a callback.  In C++, it'd be trivial to couple those settings with the callback code in a way that is thread-safe.  But in C, if the library provider makes these oversights then there basically aren't any good options, other than to write your own sort, or to basically write one callback function per pair of params.

But yeah, I agree that simplifying things should be a goal right now.  Hopefully my efforts to make Geeqie more MVC-like will also help to make things more modular and maintainable.

PR #496, opened 2017-07-09
Shuffle files around, and update translation files to match.

xsdg:
Ran regen_potfiles.sh to add missing files and correct the location of view_file files in POTFILES.in

Also, deletes a file I hadn't meant to keep.


xsdg:
Thanks!  Ran `make update-po`.

xsdg:
(Rather, I did that, then combined another change into this one, and redid it :o)

PR #498, opened 2017-07-09
Compile geeqie in C99 mode

xsdg:
Also, make AC_INIT macro independent of available branch names.


xsdg:
@gdt is probably interested in this.

PR #513, opened 2017-09-16
Fix out-of-path (VPATH) builds.

mandree:


mandree:
To reproduce the problem, do this, and you'll see that the installation fails when installing the documentation.
- run the autotools bootstrap
- mkdir _build
- cd _build
- ../configure
- make
- make install

PR #514, opened 2017-09-17
Permit reinstallation of symlink.

mandree:
This uses ln -s -f so it can overwrite the destination,
rather than fail with EEXIST.


caclark:
Implemented in commit 2d5611e5 at geeqie.org

PR #517, opened 2017-09-21
Fix checkbox falling out of sync

ilexiris:
Sync `Select > Show Marks` when it is forced to be enabled by `Select > Mark n > Filter mark n`.


caclark:
Thanks
Fixed in commit 44b347085e at geeqie.org

PR #573, opened 2018-01-03
Partially fix #561: Hi-DPI-aware rendering of images.

### LINKED ISSUE #561: Displayed images, image overlays and thumbnails are blurry on Hi-DPI displays
### ISSUE TYPE
* Bug Report

### GEEQIE VERSION

```
Geeqie 5a6cd20
```

### SUMMARY

On Hi-DPI displays, GTK/GDK application pixels automatically correspond to blocks of device pixels as described [here](https://wiki.gnome.org/HowDoI/HiDpi). In addition to querying the screen/window size (which is in application pixels), applications can also query the scale factor between device pixels and application pixels. If I understand correctly, geeqie should display higher resolution pixbufs on its UI (multiply the pixbuf size by the scale factor) in order to use the actual resolution of the display. Geeqie apparently does not do this, therefore displayed images and image overlays are blurry on Hi-DPI displays.

Additionally, in Hi-DPI mode a grid appears over the displayed images. This is probably a consequence of cutting pixbufs to tiles and displaying them side-by-side. In Hi-DPI mode these tiles do not seem to touch each other. This may be a geeqie bug or a GTK/GDK bug.

### STEPS TO REPRODUCE

One way to set up Hi-DPI mode is using the system preferences of almost any desktop environment. This usually results in a Hi-DPI scale factor of 2. This already makes the issue visible, but for testing I recommend a scale factor of 4, which makes the problem even more apparent. To do this, start geeqie by issuing `GDK_SCALE=4 geeqie`.

I attached some screenshots that demonstrate the problem (please open them in full resolution for inspection). First, for a reference, I started geeqie without any Hi-DPI scaling by issuing `GDK_SCALE=1 geeqie` and displayed a photo:

unscaled

Please note that on the screenshot above the zoom level is 1:1, so what you see are the actual pixels of the photo.

Next, I started geeqie using `GDK_SCALE=4 geeqie` and displayed the same photo:

scaled

Things worth noting here:

- The text in the status bar is very crisp, because it is handled by GDK/GTK.
- In contrast, the text in the image information overlay is very blurry. The surrounding white rectangle is blurry as well.
- The displayed photo has clearly lost a lot of detail compared to the non-HiDPI screenshot and has become very blurry. Now it consists of 4x4 blocks.
- There is a clearly visible grid over the image.

Please also note that on this screenshot the zoom level is 1:4; the photo had to be zoomed out in order to appear at the same size. Although one could argue about whether a zoom of 1:1 or 1:4 should result in the same displayed size (i.e., whether the Hi-DPI scaling factor should become a part of the zoom ratio or not), either of them would be fine on its own as long as it works consitently with fit to screen. The important thing is that since we zoomed out, 4x4 image pixels correspond to a single application pixel, which in turn corresponds to 4x4 device pixels (due to Hi-DPI scaling). So since we have an image pixel for every device pixel, we should see individual pixels as separate device pixels.

zivanfi:
Displayed images are no longer blurry on Hi-DPI displays, nor they have a grid
artifact any more.

Image overlays and thumbnails are not addresset yet and are still blurry.

### DIFF
--- a/src/renderer-tiles.c
+++ b/src/renderer-tiles.c
@@ -147,6 +147,7 @@ struct _RendererTiles
 	gint x_scroll;  /* allow local adjustment and mirroring */
 	gint y_scroll;
 
+	gint hidpi_scale;
 };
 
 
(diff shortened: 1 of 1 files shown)


caclark:
Commit 705cf22ee in main repository at geeqie.org.

PR #617, opened 2018-06-23
Spanish translation update

micrococo:


caclark:
Implemented in commit e5bfd6540 in the repo at geeqie.org.
It will be synced back here in due course.

PR #618, opened 2018-06-24
Avoid using PATH_MAX where not available

gusnan:
Geeqie fails to build on Hurd because of not finding PATH_MAX, this fix would avoid using PATH_MAX on systems where it isn't available. See https://linux.die.net/man/3/realpath .


caclark:
Implemented in commit c84ca102b9 in the repo at geeqie.org.
It will be synced back to github in due course.

PR #620, opened 2018-06-25
Fix build on x32

gusnan:
Format %lu expects 'long unsigned int' but argument is 'time_t' (aka long long int)

### DIFF
--- a/src/thumb_standard.c
+++ b/src/thumb_standard.c
@@ -348,7 +348,7 @@ static void thumb_loader_std_save(ThumbLoaderStd *tl, GdkPixbuf *pixbuf)
 		mark_uri = (tl->cache_local) ? tl->local_uri :tl->thumb_uri;
 
 		mark_app = g_strdup_printf("%s %s", GQ_APPNAME, VERSION);
-		mark_mtime = g_strdup_printf("%lu", tl->source_mtime);
+		mark_mtime = g_strdup_printf("%llu", tl->source_mtime);
 
 		pathl = path_from_utf8(tmp_path);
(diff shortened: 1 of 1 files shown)


caclark:
I can't compile with this patch. The error is:
`thumb_standard.c:351:40: error: format specifies type 'unsigned long long' but the argument has type 'time_t'
      (aka 'long') `

The gnu docs show:
`In the GNU C Library, time_t is equivalent to long int. `

I do not know how to progress this.

gusnan:
hmm, that's interesting - you can see the log for the failing build on https://buildd.debian.org/status/fetch.php?pkg=geeqie&arch=x32&ver=1%3A1.4-6&stamp=1529935765&raw=0

gusnan:
hmm, it might just be because we are missing the cast - we should probably do something like:

`mark_mtime = g_strdup_printf("%" G_GUINT64_FORMAT , (unsigned long long) tl->source_mtime);`

here I use G_GUINT64_FORMAT which simply is a platform independent variant of
llu, see the second paragraph under section "String precision pitfalls" on 
https://developer.gnome.org/glib/stable/glib-String-Utility-Functions.html

caclark:
On my system G_GUINT64_FORMAT ends up back at  "lu"

This compiles for me:

`mark_mtime = g_strdup_printf("%llu", (unsigned long long)tl->source_mtime);`

I have not been able to compile for X32 on my computer, so I can't check this.

gusnan:
Great - I just tested on x32, and that would work there too. (Sorry for being a bit of a PITA)...

caclark:
Implemented in commit b9f4412dfd at geeqie.org

PR #627, opened 2018-07-31
Add awk shebang to geocode-parameters.awk

gusnan:
The Debian lintian tools gives a warning for the geocode-parameters awk script missing a proper shebang.


caclark:
Fixed in commit fe1cddea3 at geeqie.org

PR #630, opened 2018-08-21
add headline field to the sidebar

oxpa:
Hello.

This commit adds a separate panel which allows 'headline' field to be filled.
It is needed for photo stock sites which require IPTC fields to be filled: headline, caption, etc. Geeqie had every other required field, but the headline.

Obviously, there are other ways to make 'headline' editable. Feel free to propose other approach.

### DIFF
--- a/src/bar.c
+++ b/src/bar.c
@@ -66,6 +66,15 @@ static const gchar default_config_title[] =
 "    </layout>"
 "</gq>";
 
+static const gchar default_config_headline[] =
+"<gq>"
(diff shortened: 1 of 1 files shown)


caclark:
Implemented in commit:
http://geeqie.org/cgi-bin/gitweb.cgi?p=geeqie.git;a=commit;h=74ec899812d43a3b2a348d1208404c5ce802849b

Maybe this item, along with Copyright, should be included in the Search page - but I think that will need a redesign of the page as it is getting too large.

PR #632, opened 2018-09-23
Add FILES section to manual page

gusnan:
This patch adds a FILES section to the manual page - There's a bug report on Debian which requests this section to be added to the manual page - the data I have filled it with is simply taken from 

http://www.geeqie.org/help/GuideReferenceConfig.html

### DIFF
--- a/geeqie.1
+++ b/geeqie.1
@@ -644,6 +644,59 @@ delete selected files
 close window
 .br
 
+.SH FILES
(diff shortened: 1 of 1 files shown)


caclark:
Fixed in commit:
http://geeqie.org/cgi-bin/gitweb.cgi?p=geeqie.git;a=commit;h=574ccd0bf41689b463716cacda1c061a880656ed

PR #638, opened 2018-10-06
Avoid non-portable == for the test command

Rhialto:
The test command doesn't accept `==` as operator. It should be a single `=` for portable use. The `==` is a gnu extension.

Note that the x-trick hasn't been needed for a long, long time. You can reliably write it with quotes:
```
if test "$gtk3" = yes; then
```
but I left that alone since it doesn't hurt other than being ugly.


caclark:
Fixed in commit:
http://geeqie.org/cgi-bin/gitweb.cgi?p=geeqie.git;a=commit;h=73c076b867520bd3e06e6c082f478216c2762f01

PR #647, opened 2018-11-23
rename configure.in to configure.ac

grumpytetra:


caclark:
Implemented in commit:
http://geeqie.org/cgi-bin/gitweb.cgi?p=geeqie.git;a=commit;h=939909993f949d2af5e8bbf9970e5510a5d4742e

PR #648, opened 2018-11-26
Correctly size image windows opened from collections

davelambert:
When opening a view window from a collection, get the the image size
info from the file so the new window is appropriately sized.


caclark:
Implemented in commit:
http://geeqie.org/cgi-bin/gitweb.cgi?p=geeqie.git;a=commit;h=d3c0882aeed682feebd59946229c55ab02b70ea1

PR #649, opened 2018-11-26
Tell Git to ignore .desktop files globally

davelambert:
These files are all generated by the build, with the exception of the
template.desktop file, so:

* Have Git ignore all *.desktop files.
* Rename template.desktop to template.desktop.in to
  avoid a clash with this new ignore rule.
* Remove the checked-in image-crop.desktop file, which is built
  anyway.


caclark:
Hello Dave
Thanks for the patch, but I think the file template.desktop is not being created. When I tried this patch it looked like the template.desktop being used was from an earlier install.

davelambert:
Hi Colin,

Oops!  I missed the use of desktop.template at runtime in the plugin configuration window.  I've changed the commit so DESKTOP_FILE_TEMPLATE references the .in version (creating a new plugin now shows the contents in the editor for me), if you would like to try again.

Thanks,
Dave

caclark:
Hello Dave
I changed your patch slightly to include template.desktop as part of the language translation process.
Well, actually the comments in the desktop file need to be translated, but I don't know how to achieve that. Just another unsolved problem.
http://geeqie.org/cgi-bin/gitweb.cgi?p=geeqie.git;a=commit;h=969791e9521aeffd61d67b3c2d17ccbff98f00d6

PR #651, opened 2018-12-07
Fix fullscreen spanning multiple monitors

### LINKED ISSUE #650: fullscreen does not work over multiple monitors
### ISSUE TYPE
* Bug Report
### GEEQIE VERSION
geeqie 1.3 and geeqie_1.4+git20181019
### OS / DISTRIBUTION
Debian stable
Problem in both the distribution version of geeqie and self-backported package from sid.
### SUMMARY
The system has two monitors forming a single wide screen spanning both monitors, set up with xrandr.
In Debian Jessie, geeqie was able to do fullscreen spanning both monitors.
Now, only one monitor is used for Fullscreen. We use geeqie for polarized stereo projection and cannot use it after updating to Debian Stretch and beyond.
My window manager is icewm, but I get the same results with kwin, metacity or openbox.
However, when I don't run any window manager, the fullscreen works correctly over both monitors.

### STEPS TO REPRODUCE
1) Set up two monitors with something like
xrandr --output VGA-1 --auto
xrandr --output HDMI-1 --auto --right-of VGA-1

2) go to Preferences->Windows->Fullscreen and in Location select either "Active screen" or "Screen 0:0, full size"

3) press "f" for full screen.

Image should be displayed on both monitors, but it is not.

I believe this is caused by commit a16635ca1d8fe42349f9e22bd311d3118eff2608 ("Fix fullscreen - remove all the garbage which breaks things and let Gtk do its job.")
I have spent some time looking for ways to fix this. I think gtk_window_fullscreen() will not use multiple monitors without some extra work. gdk-window-set-fullscreen-mode() has been introduced in GDK 3.8 which seems to do just what is needed. 

I see three possibilities
1) backport gdk-window-set-fullscreen-mode() into src/compat.c to get that functionality without introducing a dependency on GDK 3.8+
2) use low level X11 magic directly, as implemented by gdk_x11_window_apply_fullscreen_mode(). googling "_NET_WM_FULLSCREEN_MONITORS" shows many projects using this
3) bring back some of the code removed by @jkufner  in a16635ca1d8fe42349f9e22bd311d3118eff2608.

Indeed using just gtk_window_fullscreen() without any further hacks gives better result for single-monitor fullscreen.
However, when one of the "Screen *, full size" options is chosen, i.e. "options->fullscreen.screen % 100 == 0", we could use the old
code, making use of gdk_window_set_override_redirect() and gtk_window_set_keep_above() instead of gtk_window_fullscreen()

I can try to fix this, but being very new, I would like to get some advice which option seems most acceptable.

Thanks for any hints!

jiribohac:
Commit a16635c ("Fix fullscreen - remove all the garbage
which breaks things and let Gtk do its job.")
broke fullscreen spanning multiple monitors.

Fix this for GTK3 by using gdk_window_set_fullscreen_mode().
Fixes #650.

Unfortunately, GTK2 has no equivalent.

Tested with kwin and icewm.
BTW, I found that in Icewm, selecting different monitors for fullscreen has no effect - with or without this patch. I'll try to debug that separately.

### DIFF
--- a/src/fullscreen.c
+++ b/src/fullscreen.c
@@ -248,13 +248,6 @@ FullScreenData *fullscreen_start(GtkWidget *window, ImageWindow *imd,
 	gtk_window_set_decorated(GTK_WINDOW(fs->window), FALSE);
 	gtk_container_set_border_width(GTK_CONTAINER(fs->window), 0);
 
-	/* make window fullscreen -- let Gtk do it's job, don't screw it in any way */
-	gtk_window_fullscreen(GTK_WINDOW(fs->window));
-
-	/* move it to requested screen */
-	if (options->fullscreen.screen >= 0) {
-		gtk_window_set_screen(GTK_WINDOW(fs->window), screen);
-	}
 
 	/* keep window above others, if requested */
 	if (options->fullscreen.above) {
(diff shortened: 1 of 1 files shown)


caclark:
Implemented in commit:
http://geeqie.org/cgi-bin/gitweb.cgi?p=geeqie.git;a=commit;h=1d8d758e8adc56e1ff9bf0defccaa47bcb08c3b4

PR #658, opened 2019-01-20
Bug fix: reads of uninitialized data

i-rinat:
There are a couple places where variables can be used without prior initialization.
`n` from `exif_build_formatted_star_rating()` is quite harmless, but `link` in `vdlist_populate()` being then passed to `free()` can cause a double-free, or a use-after-free.


i-rinat:
...Can and actually does. Geeqie often crashes when I open a directory with many images.

caclark:
Fixed in commit:
http://geeqie.org/cgi-bin/gitweb.cgi?p=geeqie.git;a=commit;h=431adf320f4ce0ba31fb7a07ea53aca03c21fcbd

PR #662, opened 2019-02-26
Fix preferences dialog file tab filter crash (Closes #661)

### LINKED ISSUE #661: String comparison to translated string causes crash
### ISSUE TYPE
* Bug Report

### GEEQIE VERSION
Latest Git version of Geeqie, GTK 2

### OS / DISTRIBUTION
Debian Testing / unstable

### SUMMARY
Running Geeqie in non-English (or to be specific, in languages where
"Filter" is translated to something other than "Filter") crashes geeqie
when trying to add a new filter in the Files tab of the Preferences dialog.

See the check in src/preferences.c line 1065, where the code compares a
column title to "Filter", without taking translations into consideration,
which means the comparison will fail on any language where "Filter" is
translated to something other than "Filter".

I couldn't reproduce the problem in the Debian bug report, but that is because
I run in Swedish, and the word "Filter" is the same in English as in Swedish.

See the Debian bug report at
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=923018

gusnan:
Instead of searching for the column where the title is
"Filter", simply use column 0, which is already done in
several other places in the code. This fixes the problem
where not finding the proper column in languages where the
column hasn't got the title "Filter".

### DIFF
--- a/src/preferences.c
+++ b/src/preferences.c
@@ -1049,7 +1049,6 @@ static gboolean filter_add_scroll(gpointer data)
 	GList *list_cells;
 	GtkCellRenderer *cell;
 	GtkTreeViewColumn *column;
-	GList *list_columns;
 	const gchar *title;
 	guint i = 0;
 	gint rows;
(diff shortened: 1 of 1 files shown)


caclark:
Thanks

Fixed in commit:
http://geeqie.org/cgi-bin/gitweb.cgi?p=geeqie.git;a=commit;h=0004617a71100553a13e49e7208e401f2493b72c

jEsuSdA:
Thanks a lot!!! ;)

PR #703, opened 2019-08-02
Fix build on non-glibc platforms

AMDmi3:
`__GLIBC_PREREQ` is not defined on non-glibc platforms (for instance, on FreeBSD), build fails there. So switch to checking glibc version only after checking for that it's available. Also enable strftime extensions on FreeBSD which does support them.

### DIFF
--- a/src/pan-view/pan-util.c
+++ b/src/pan-view/pan-util.c
@@ -82,6 +82,12 @@ gint pan_date_value(time_t d, PanDateLengthType length)
 	return -1;
 }
 
+#if defined(__GLIBC_PREREQ)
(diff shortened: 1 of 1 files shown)


AMDmi3:
Would be even better to add configure check instead of ifdef construct, but I'm not into autotools enough to implement it myself.

caclark:
Implemented in main repo. at geeqie.org:
http://geeqie.org/cgi-bin/gitweb.cgi?p=geeqie.git;a=commit;h=d09b06c05c61cc675726a440f9cc787807024bb8

PR #742, opened 2020-01-07
Add appdata file

### LINKED ISSUE #373: Please consider committing and installing the supplied AppData file
Please consider installing this AppData file: https://raw.githubusercontent.com/hughsie/fedora-appstream/master/appdata-extra/desktop/geeqie.appdata.xml

This is used in GNOME and KDE software installers to add the application description and some screenshots. We'd love to showcase more applications, but without the extra data file we can't. The AppData file needs to be installed to /usr/share/appdata/ on Linux and the basename needs to match the .desktop basename. It would also be great if you could integrate the file with your translation system (e.g. intltool) to make the descriptions translated.

See http://people.freedesktop.org/~hughsient/appdata/ for more details; thanks!

Richard

Reported by: hughsie

Hofer-Julian:
Fixes #373

### DIFF
--- a/org.geeqie.Geeqie.appdata.xml
+++ b/org.geeqie.Geeqie.appdata.xml
@@ -1,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component>
+  <id type="desktop">geeqie.desktop</id>
+    <metadata_license>CC-BY-SA-4.0</metadata_license>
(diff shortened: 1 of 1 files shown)


caclark:
Implemented in commit:
http://www.geeqie.org/cgi-bin/gitweb.cgi?p=geeqie.git;a=commit;h=0fe09634043fb78aca550f6657923ffb44b761bd

PR #746, opened 2020-01-23
src/options: fix build failure against gcc-10

trofi:
On gcc-10 (and gcc-9 -fno-common) build fails as:

```
  CXXLD    geeqie
ld: ui_fileops.o:(.bss+0x0):
  multiple definition of `options'; ui_bookmark.o:(.bss+0x0): first defined here
ld: ui_fileops.o:(.bss+0x8):
  multiple definition of `command_line'; ui_bookmark.o:(.bss+0x8): first defined here
```

gcc-10 will change the default from -fcommon to fno-common:
https://gcc.gnu.org/PR85678.

The error also happens if CFLAGS=-fno-common passed explicitly.

Reported-by: Anthony Parsons
Bug: https://bugs.gentoo.org/706132


caclark:
Thanks. Implemented in commit:
http://geeqie.org/cgi-bin/gitweb.cgi?p=geeqie.git;a=commit;h=3ea5fa39d37e56443907255fbc81cb1b83dc9978

PR #752, opened 2020-01-30
Allow scroll bookmark window horizontally

przepompownia:


przepompownia:
Note that the current github clone is more than a half year behind git://www.geeqie.org/geeqie.git

caclark:
I think this is the relevant part of the .diff:
--- a/src/ui_bookmark.c
+++ b/src/ui_bookmark.c
@@ -806,7 +806,7 @@ GtkWidget *bookmark_list_new(const gchar *key,
 
 	scrolled = gtk_scrolled_window_new(NULL, NULL);
 	gtk_scrolled_window_set_policy(GTK_SCROLLED_WINDOW(scrolled),
-				       GTK_POLICY_NEVER, GTK_POLICY_AUTOMATIC);
+				       GTK_POLICY_AUTOMATIC, GTK_POLICY_AUTOMATIC);

The first image here is of the current situation,and I assume that is what you are trying to avoid.
The second image is with your patch, and that also has problems (on my system).
Maybe the following code could be included:
`PangoLayout *layout = gtk_widget_create_pango_layout(GTK_WIDGET(scrolled), "this is reasonable width");

gint width,height;

pango_layout_get_pixel_size(layout, &width, &height);

gtk_scrolled_window_set_min_content_width(GTK_SCROLLED_WINDOW(scrolled), width);
`
This is GTK3 only. I also don't know if this is valid for 4K monitors. I am not at home at the moment so I cannot check that.
Screenshot from 2020-01-31 12-12-21
Screenshot from 2020-01-31 12-14-17

przepompownia:
Yes, sometimes I need to drag the right border of the bookmark view at the cost of directory view space to see longer names (in practice it may be name about ten characters).

I am not familiar with any GTK and C/C++. So, as you can expect, it is a hack that works for me and require tests and more work for some other cases (including GTK2).

caclark:
Fixed in commit:
http://geeqie.org/cgi-bin/gitweb.cgi?p=geeqie.git;a=commit;h=d1bd7992378a680ac81f9fc151ec6fa60b34eba3

PR #753, opened 2020-02-04
Various cleanups based on gcc warnings

keszybz:
Note that those patches are based on the master branch on geeqie.org. Please update the master branch on github too.


caclark:
Would you please post a .diff against the repository at geeqie.org?
Thanks, and also include a list of gcc options you used to get the warnings.

Post #752 also mentions that the GitHub repo. is out of date.
GitHub is very useful for tracking the ever-increasing number of bugs and feature requests.
But for me, it is a pain in the neck - I would rather that geeqie.org was the only one to be cloned from.
My opinion is that if it cannot be auto synced. from geeqie.org, it should be ignored.

caclark:
Thanks.
Implemented in commit:
http://geeqie.org/cgi-bin/gitweb.cgi?p=geeqie.git;a=commit;h=95877d468a2b9a52d8104ec2698418d7e5e1309e

PR #759, opened 2020-03-11
NetBSD fixes

0-wiz-0:
`sys/sysinfo.h` does not exist on NetBSD, but nothing seems to be used from it.
(If it is useful on other system, please autodetect it, or wrap it in `#ifndef __NetBSD__`.)

`sh` is sufficient for `gen_changelog.sh` and `/bin/bash` does not exist on NetBSD (you can install it, but then it's in `/usr/pkg/bin/bash` by default).

`sys/wait.h` is needed to get the `WEXITSTATUS` symbol.

### DIFF
--- a/gen_changelog.sh
+++ b/gen_changelog.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 # Script to update ChangeLog file,
 # it keeps "pre-svn" history and inserts git log at top,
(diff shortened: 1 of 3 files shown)


caclark:
Fixed in commit:
http://geeqie.org/cgi-bin/gitweb.cgi?p=geeqie.git;a=commit;h=701d7c379bcb90feb7bea07c2355207b21a8f105

PR #771, opened 2020-05-19
properly check return value of mmap

hannob:
This check for the return value of the mmap function is wrong.

mmap returns MAP_FAILED (or -1) on errors, so the "if(!libary->mapping)" clause will never fail and the error code will not be called on failure of mmap.

### DIFF
--- a/src/zonedetect.c
+++ b/src/zonedetect.c
@@ -503,7 +503,7 @@ ZoneDetect *ZDOpenDatabase(const char *path)
         lseek(library->fd, 0, SEEK_SET);
 
         library->mapping = mmap(NULL, (size_t)library->length, PROT_READ, MAP_PRIVATE | MAP_FILE, library->fd, 0);
-        if(!library->mapping) {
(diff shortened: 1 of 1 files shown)


caclark:
This code is taken from Bertold Van den Bergh's ZoneDetect project on GitHub:
https://github.com/BertoldVdb/ZoneDetect

I simply made a copy of the files from that project, as I did not know any other way to incorporate it.

It would be better for the change to be made first in the ZoneDetect project.
Would you like to contact Bertold Van den Bergh, or would you prefer me to do that?

hannob:
I'll do a pullrequest for zonedetect.

hannob:
ZoneDetect has merged the change, so if you re-sync with their code this issue is resolved.

caclark:
Thanks.
Fixed in commit:
http://geeqie.org/cgi-bin/gitweb.cgi?p=geeqie.git;a=commit;h=d9ecec6e4cf3f8121aac7efb4aeda0fa2cb5e4c2

PR #779, opened 2020-06-22
Add search by path

gusr:
Added the possibility to search for files by path in the search dialog. This is essential if, for instance, you have a directory tree containing plots generated from a set of experiments, such that the directory names along the path describe the experimental parameters, and the file name describes a type of visualization.

I chose to implement this as an alternative form of the file name search option. This avoids adding another row to the UI, and also minimizes the amount of code added. (If a separate field for path had been added instead that would also only make sense when the "recurse" box is checked.)

### DIFF
--- a/src/search.c
+++ b/src/search.c
@@ -69,9 +69,6 @@ typedef enum {
 	SEARCH_MATCH_NONE,
 	SEARCH_MATCH_EQUAL,
 	SEARCH_MATCH_CONTAINS,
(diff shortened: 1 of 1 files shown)


caclark:
Implemented in commit:
http://geeqie.org/cgi-bin/gitweb.cgi?p=geeqie.git;a=commit;h=99aa4caeef1f7c2985431e3e15f95563a30fb354

PR #808, opened 2020-09-19
Update Slovak translation

jose1711:


caclark:
Implemented in commit:
http://geeqie.org/cgi-bin/gitweb.cgi?p=geeqie.git;a=commit;h=8bdbba90033f966fdd70acb1a1324fa9e843ee0d

PR #810, opened 2020-09-20
Fix mistakes in Czech translation

jose1711:


caclark:
Implemented in commit:
http://geeqie.org/cgi-bin/gitweb.cgi?p=geeqie.git;a=commit;h=266f15fc0b01a46b45011338ffe5492b6cde74b0

PR #816, opened 2020-10-03
Fix building with --enable-debug-flags

### LINKED ISSUE #815: The --enable-debug-flags option results in compile errors
### ISSUE TYPE

* Bug Report

### GEEQIE VERSION

```
Geeqie 1.5.1+git20200904-0ff5c85b

```

### STEPS TO REPRODUCE

Execute:
`./autogen.sh  --enable-debug-flags`
`make`

gusnan:
I have made something that should fix #815 , but note that this pull-request includes several other commits that was only on the geeqie git repository, but not here. This makes it harder to work out. See "Fix building with --enable-debug-flags" below.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -77,3 +77,4 @@ Makefile.in
 ChangeLog.html
 README.html
 ChangeLog
+/org.geeqie.Geeqie.appdata.xml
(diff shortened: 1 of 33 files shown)


caclark:
Thanks. (I changed a couple of function names to avoid name conflicts with gtk libraries).

Implemented in commit:
http://geeqie.org/cgi-bin/gitweb.cgi?p=geeqie.git;a=commit;h=a54864bf069d0d3c7f6dd1ba625416e365124fa1

PR #831, opened 2020-12-06
exclude remember workspace

yeezy69:
Please add a new, separate setting option so that geeqie does not remember the current workspace, if window positions/sizes are saved. 

I work with several workspaces. It is very ugly while opening geeqie it jumps to a different/previous workspace than the current workspace.

Thank you!


caclark:
Fixed in commit:
http://geeqie.org/cgi-bin/gitweb.cgi?p=geeqie.git;a=commit;h=9443a0929a6f73010f5f86c3147ada7288a10f8a

PR #882, opened 2021-04-09
Fix doc building with newer yelp

mattdm:
This adds `-i`, because otherwise it will complain about missing files (which
we copy into place later), and adds a trailing slash to the output directory,
because `yelp-build` now puts the files into the wrong place otherwise.


caclark:
Implemented in commit a33d8c40

PR #885, opened 2021-04-26
Generate sRGB thumbnails for images with embedded color profile data

### LINKED ISSUE #608: Thumbnails not color-managed
`v1.2-476-g`7ce4193d
Save the images from http://www.color.org/version4html.xalter
The images display correctly in the main area, but the thumbnails seem not to be color-managed at all (both ICC v2 and v4 image thumbs are incorrect).

cytrinox:
Images with color profiles other than sRGB or AdobeRGB may look too bright/dark as thumbnails, because Geeqie ignores color profile information during thumbnail generation.

This patch adds color profile correction for thumbnails if the image contains EXIF data and provides an embedded color profile or EXIF color profile identifiers.

Fixes #608

### DIFF
--- a/src/thumb.c
+++ b/src/thumb.c
@@ -145,6 +145,11 @@ static void thumb_loader_done_cb(ImageLoader *il, gpointer data)
 		return;
 		}
 
+	if(!tl->cache_hit)
+		{
+			// apply color correction, if required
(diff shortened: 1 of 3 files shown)


caclark:
There is an additional option in `Edit/Preference/General` to select thumbnail color management.
The default is "off".
If `Use EXIF thumbnails` is on, this option will be ignored.

PR #895, opened 2021-05-15
Update Slovak translation

jose1711:


caclark:
Implemented in commit d1f51506579

PR #898, opened 2021-05-26
fixed ubuntu url

lrkwz:


caclark:
Implemented in commit 163d3eeb

PR #908, opened 2021-06-26
Sk translation

jose1711:


caclark:
Thanks but the .diff and .patch are significantly different in size - so I assume the .patch is the correct one.
But I get an error when trying to apply it:
`Applying: Update Slovak translation ` 
`error: patch failed: po/sk.po:10`
`error: po/sk.po: patch does not apply`
`Patch failed at 0001 Update Slovak translation`

jose1711:
> Thanks but the .diff and .patch are significantly different in size - so I assume the .patch is the correct one.
> But I get an error when trying to apply it:
> `Applying: Update Slovak translation `
> `error: patch failed: po/sk.po:10`
> `error: po/sk.po: patch does not apply`
> `Patch failed at 0001 Update Slovak translation`

Sorry, my bad. I rebased onto current master - is it better now?

caclark:
Implemented in commit febb7904

PR #921, opened 2021-09-04
Fix spelling error avaialble/available

gusnan:
Fix spelling error avaialble/available in preferences.c


caclark:
Fixed in commit 131a42a

PR #931, opened 2021-10-22
Simple maintenance improvements

a1346054:
Review comments are very welcome, I can fixup commits as needed.


caclark:
Thanks for the work.
Implemented in 717f743 et al

PR #937, opened 2021-11-26
Add content to the generic dialogs window

floemker:
In case an older GTK version is used the function generic_dialog_setup()
did not add it's content to the dialog window.


caclark:
Pushed in commit: f6bc1d2ead

PR #938, opened 2021-12-07
Fix building whereami on Hurd/kFreeBSD

gusnan:
Fixes building geeqie and it's embedded whereami on Hurd (and also kFreeBSD) - One option would be to not embed it in geeqie, but use distros provided packages, where this patch is already applied - at least in Debian it is available:

https://packages.debian.org/sid/libwhereami0

https://salsa.debian.org/yangfl-guest/whereami/-/blob/master/debian/patches/0001-added-support-for-kFreeBSD-and-GNU-Hurd.patch

### DIFF
--- a/src/whereami.c
+++ b/src/whereami.c
@@ -3,6 +3,10 @@
 //   by Gregory Pakosz (@gpakosz)
 // https://github.com/gpakosz/whereami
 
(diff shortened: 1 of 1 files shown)


caclark:
Pushed in commit: 7277fda2

PR #942, opened 2021-12-15
Add image/heic MIME type

dfandrich:


caclark:
Pushed in commit: a0598407

PR #951, opened 2022-01-05
Add menu item to open containing folder

gusnan:
I was looking for a method to open the containing folder in my prefered file browser, and couldn't find anything, so I implemented this - it might be implemented in more (other) places, but for my needs, this is enough.

Please comment if it needs changes.

### DIFF
--- a/src/view_file/view_file.c
+++ b/src/view_file/view_file.c
@@ -611,6 +611,28 @@ static void vf_pop_menu_collections_cb(GtkWidget *widget, gpointer data)
 	filelist_free(selection_list);
 }
 
+static void open_folder_cb(GtkWidget *widget, gpointer data)
+{
+	ViewFile *vf = data;
+	GError *err = NULL;
+
+	GList *list = vf_pop_menu_file_list(vf);
+
+	GList *work = list;
+
+	// Get the first item in the list
+	FileData *fd = work->data;
+
+	if (fd) {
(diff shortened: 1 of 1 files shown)


caclark:
I think this function can be also achieved by a plug-in.
The advantage of a plug-in is that you can set a hot-key or place an icon on the toolbar. It will also run wherever (maybe) the focus is.
The disadvantage is that some people may prefer the right-click style of working.

It is also possible to place the plug-in on any of the main menus. For a little information about that go to `Help/On-Line Help Search` and look for `plugins configuration`.

gusnan:
Thank you - you are indeed right, this might be better as a plugin - I have managed to make something that I can use. One small question - is there a way to get the plugin menu item in the right-click menu? From what you say so far, I believe it cannot be done, but as you also say, giving it a keyboard shortcut might be equally good for me.

caclark:
The pop-up menus are all hard-coded in various parts of the project. They are a pain in the neck, because some overlap with the main menu functions but are handled separately. There is no way to change this.
Maybe there will be a possibility with GTK4, but that is so far away it is rather irrelevant at the moment

gusnan:
Thanks - I am satified with using a plugin, which makes it possible to add a keyboard shortcut. Feel free to close this bug/PR. Thanks for your work on geeqie!

PR #961, opened 2022-01-29
Update Slovak translation

jose1711:


caclark:
Fixed in commit ea678d19

PR #974, opened 2022-03-09
Fix comparison in version.sh

AMDmi3:
Though some `test` implementations do support `==` for compatibility, standard primary is `=`


caclark:
Thanks. Fixed in commit 3ffa1b284

PR #975, opened 2022-03-09
Fix version.sh shebang

AMDmi3:
This script does not require bash to run, and will in fact fail to run on systems without bash

### DIFF
--- a/version.sh
+++ b/version.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 ## @file
 ## @brief Generate the Geeqie version number


caclark:
Unfortunately all of the scripts I have created use bash. Is this a problem for FreeBSD? Should they all be converted to POSIX compliant?

AMDmi3:
Well because scripts require bash I have to pull it as a dependency for FreeBSD port of geeqie and fix the shebangs (FreeBSD doesn't have bash by default), so I'd definitely prefer scripts to be POSIX-compatible instead, unless using bash significantly simplifies the development. This script though is separate from the installed ones and it doesn't seem to use anything bash specific so it can be fixed right away.

PR #977, opened 2022-03-11
Use sh instead of bash for build scripts

ptrcnull:
These scripts are POSIX sh compatible anyway, so changing the hashbang to `#!/bin/sh` avoids having to pull in unnecessary dependencies.


caclark:
Thanks. Fixed in commit 3ffa1b284b

PR #980, opened 2022-04-01
Add AVIF file extension to filefilter.c

cytrinox:
AVIF is supported by geeqie via libheif, but files are filtered out because the extension is not registered yet.

### DIFF
--- a/src/filefilter.c
+++ b/src/filefilter.c
@@ -296,6 +296,7 @@ void filter_add_defaults(void)
 #endif
 #ifdef HAVE_HEIF
 	filter_add_if_missing("HEIF", "HEIF Format", ".heic;.heif", FORMAT_CLASS_IMAGE, FALSE, FALSE, TRUE);
+	filter_add_if_missing("AVIF", "AVIF Format", ".avif", FORMAT_CLASS_IMAGE, FALSE, FALSE, TRUE);
 #endif
 #ifdef HAVE_WEBP
 	filter_add_if_missing("webp", "WebP Format", ".webp", FORMAT_CLASS_IMAGE, FALSE, FALSE, TRUE);


caclark:
Implemented in commit 71504c019c9

tomaszg7:
This pull request makes a mess in the filter. If `gdk-pixbuf` already supports AVIF/HEIF we get double entries with different flags:

```
            <file_type key = "heif/avif" enabled = "true" extensions = ".heif;.heic;.avif" description = "HEIF/AVIF Image" file_class = "1" writable = "true" allow_sidecar = "false" />
            <file_type key = "HEIF" enabled = "true" extensions = ".heic;.heif" description = "HEIF Format" file_class = "1" writable = "false" allow_sidecar = "false" />
            <file_type key = "AVIF" enabled = "true" extensions = ".avif" description = "AVIF Format" file_class = "1" writable = "false" allow_sidecar = "false" />
```

Also, why is sidecar not allowed here? Any reason for this?

To add to the mess, added keys are for some reason uppercase, as opposed to all other keys. 

Regretfully, even fixing this code may leave users with multiple entries in their config files as Geeqie doesn't clean up the entries if they get changed.

caclark:
The mime key `heif/avif` originates in the gdk pixbuf loader in the libheif library. Although it is part of the libheif library it is installed via another package `heif-gdk-pixbuf` which at the moment is not listed as a dependency for Geeqie.
Even if it is installed the Geeqie code ignores it and uses `image_load_heif.c` as before.
I changed the code to use  `heif-gdk-pixbuf`, but that does not work -  a black image is displayed.

tomaszg7:
It is installed by another package in some distributions, in other distributions it is installed together with libheif. I didn't try to use this gdk loader. The point was that it's existence interferes with default values of the filter. Alternate solution would be to completely ignore avif/heif when querying for gdk supported formats.

PR #995, opened 2022-05-31
A few changes to filefilter.c

tomaszg7:
Several small cleanups of `filefilter.c` and one logic update. It amends [pull/980](https://github.com/BestImageViewer/geeqie/pull/980) and other older commits.

1. Change HEIF and AVIF default values to correspond to what gdk-pixbuf reports, so there are no duplicate entries.
2. Change HEIF and AVIF to `not writable` as Exiv2 doesn't support writing BMFF tags (yet?)
3. Change WEBP files to `writable`.
4. Merge entries for older Canon raw files (crw and cr2) with newer one (cr3). No reason to keep them separate here, even though cr3 has a separate loader.
5. Move formats supported by custom loaders to the front so that proper flags get applied, not default writable,no-sidecar which is used for gdk-pixbuf. Note that cusom loaders have priority in `image-load.c`.

### DIFF
--- a/src/filefilter.c
+++ b/src/filefilter.c
@@ -295,8 +295,7 @@ void filter_add_defaults(void)
 	filter_add_if_missing("pdf", "Portable Document Format", ".pdf", FORMAT_CLASS_DOCUMENT, FALSE, FALSE, TRUE);
 #endif
 #ifdef HAVE_HEIF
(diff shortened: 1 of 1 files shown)


caclark:
Implemented in commits d962f56 571aff46714 59c685be 9a1e251

PR #1000, opened 2022-07-14
Drop unconditional Xlib dependency

ya-isakov:
It was introduced 4eb2f5880956c8259d0b5e2084baf61a27ee82ad, to support XInitThreads, but XInitThreads was dropped later. Nothing else in this file needs this header, but it fails to compile on X11-less systems


caclark:
Implemented in commit 3a1509b0

PR #1007, opened 2022-07-25
Allows `meson setup` to finish successfully without champlain and without docs utils

xsdg:
As originally written, the build would fail if libchamplain weren't found (with an unknown variable when `src/meson.build` tried to access `clutter_dep` or `clutter_gtk_dep`), or if `yelp-build` or `pandoc` weren't found.  These two changes allow the build system setup to complete in either of those cases.


caclark:
Thanks - fixed in commit db95a097 - I hope.

PR #1008, opened 2022-07-25
Stop relying on implicit conversion from TRUE to COLLECTION_LOAD_APPEND.

xsdg:
This is one of the fixes that was necessary for C++ compatibility.


caclark:
Fixed in commit bb3067d5

PR #1009, opened 2022-07-25
Avoids narrowing conversion by properly declaring 32-bit masks as uint instead of int.

xsdg:
Another fix that was important for C++ compatibility.


caclark:
Applying the patch gives me a few compiler warnings, e.g.:

../src/image_load_dds.c:137:117: warning: comparison of integers of different signs: 'int' and 'const uint' (aka 'const unsigned int') [-Wsign-compare]
                        if (redMask == A1R5G5B5_MASKS[0] && greenMask == A1R5G5B5_MASKS[1] && blueMask == A1R5G5B5_MASKS[2] && alphaMask == A1R5G5B5_MASKS[3]) {
                                     

Maybe this would be enough?:
[1009-1.diff.gz](https://github.com/BestImageViewer/geeqie/files/9200216/1009-1.diff.gz)

xsdg:
Thanks for pointing those out.  It's interesting that they show up when compiled as C, but not when compiled as C++.

Either way, it's clear that this file broadly uses the `int` type for 32-bit unsigned values, so I've gone through and changed `int` to `uint` where it seemed to match the intent of the code.

Do you know where to find a sample input file to at least make sure this doesn't blow up immediately?

caclark:
This is a test image I use - origin unknown:
[test-dxt5.dds.gz](https://github.com/BestImageViewer/geeqie/files/9203064/test-dxt5.dds.gz)

xsdg:
Thanks for the sample!  I think this PR is good to go.
# baseline (commit `f02c9285`)
base_geeqie

# with the fix_narrow commits
geeqie_with_fix_narrow

caclark:
Fixed in commits f096cb15 54301aa11 92c834c3b

PR #1011, opened 2022-07-27
Fix HAVE_FFMPEGTHUMBNAILER_METADATA to use mesondefine

gusnan:
Looks to me that also HAVE_FFMPEGTHUMBNAILER_METADATA should use #mesondefine - On older distros with ffmpeg < 2.2 the video_thumbnailer_struct doesn't include the prefer_embedded_metadata field, and this causes a build error.

(I'm on libffmpegthumbnailer4v5 2.1.1-0.2+b1 on Debian 11.4).


caclark:
Fixed in commit 5f07e15

PR #1012, opened 2022-07-27
No longer attempts to run yelp-build from the docs build if it wasn't found previously.

xsdg:
This was the sole fix from #1007 that wasn't already handled in https://github.com/BestImageViewer/geeqie/commit/db95a09715494db519f47315e970cf602cb97a88


caclark:
Fixed in commit 599b716

PR #1013, opened 2022-07-28
Fixes jumps across variable declarations in same scope

xsdg:
Those jumps are not legal in C++.  zonedetect.h is an imported library, so this is a minimal change that allows compilation as C++.

### DIFF
--- a/src/zonedetect.c
+++ b/src/zonedetect.c
@@ -593,12 +593,12 @@ float* ZDPolygonToList(const ZoneDetect *library, uint32_t polygonId, size_t* le
     uint32_t polygonIndex;
     int32_t* data = NULL;
     float* flData = NULL;
+    size_t length = 0;
 
     if(!ZDFindPolygon(library, polygonId, NULL, &polygonIndex)) {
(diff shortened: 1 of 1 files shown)


caclark:
I get two more compiler warnings - I hope they are also easy to fix:

[161/177] Compiling C object src/geeqie.p/zonedetect.c.o
../src/zonedetect.c:1219:19: warning: comparison of integers of different signs: 'int' and 'unsigned long' [-Wsign-compare]
    for(int i=0; i<sizeof(strings)/sizeof(char*); i++) {
                 ~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../src/zonedetect.c:1238:23: warning: comparison of integers of different signs: 'int' and 'unsigned long' [-Wsign-compare]
        for(int i=0; i<sizeof(strings)/sizeof(char*); i++) {

xsdg:
Okay, compile warnings should be fixed.

caclark:
Fixed in commits d7cae8cd2 d3e11814e

PR #1015, opened 2022-07-31
Catalan translation updated

cpina:
Let me know if you prefer translations submitted in a different way.


caclark:
Fixed in commits 3726a40f 879833e1 2da64a2b 399ec7c32

PR #1017, opened 2022-08-01
Use source root and not build root to install Changelogs and readme from

gusnan:
This fixes a build error when using Debian chroot:

```
Run-time dependency lua5.3 found: YES 5.3.6
Run-time dependency poppler-glib found: YES 22.02.0
Run-time dependency gspell-1 found: YES 1.11.1
Library libtiff found: YES
Checking for function "TIFFClientOpen" with dependency -llibtiff: YES 
Run-time dependency libwebp found: YES 1.2.2
Checking if "nl_langinfo and _NL_TIME_FIRST_WEEKDAY" : links: YES 
Configuring config.h using configuration
Program msgfmt found: YES (/usr/bin/msgfmt)
Program msginit found: YES (/usr/bin/msginit)
Program msgmerge found: YES (/usr/bin/msgmerge)
Program xgettext found: YES (/usr/bin/xgettext)
Found pkg-config: /usr/bin/pkg-config (0.29.2)
Program glib-genmarshal found: YES (/usr/bin/glib-genmarshal)
Program generate-ClayRGB1998-icc-h.sh found: YES (/build/geeqie-1.7.3+git20220730~f02c9285/scripts/generate-ClayRGB1998-icc-h.sh)
Program ln found: YES (/bin/ln)
Program doxygen found: NO
Message: source_root: /build/geeqie-1.7.3+git20220730~f02c9285
Message: build_root: /build/geeqie-1.7.3+git20220730~f02c9285/obj-x86_64-linux-gnu
Message: current_build_dir: /build/geeqie-1.7.3+git20220730~f02c9285/obj-x86_64-linux-gnu
Program gen_changelog.sh found: YES (/build/geeqie-1.7.3+git20220730~f02c9285/gen_changelog.sh)
Program pandoc found: YES (/usr/bin/pandoc)
Program gen_readme.sh found: YES (/build/geeqie-1.7.3+git20220730~f02c9285/gen_readme.sh)

../meson.build:489:4: ERROR: File /build/geeqie-1.7.3+git20220730~f02c9285/obj-x86_64-linux-gnu/ChangeLog does not exist.

A full log can be found at /build/geeqie-1.7.3+git20220730~f02c9285/obj-x86_64-linux-gnu/meson-logs/meson-log.txt
	cd obj-x86_64-linux-gnu && tail -v -n \+0 meson-logs/meson-log.txt
```

I don't run into the error when doing a standard build, but I only run into the problem when building using Debian tools which uses a chroot and sets up a build system to make sure it's a "clean" system building on.

I guess this is the simplest fix, but I would very much appreciate if someone more knowledgeable in meson to have a look at the patch and check if it got any problems.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -483,15 +483,15 @@ if pandoc.found()
     run_command(find_program('gen_readme.sh'), meson.source_root(), meson.build_root(), check : false)
 
     install_data('README.md', 'COPYING', 'TODO', 'AUTHORS',
-                 join_paths(meson.build_root(), 'ChangeLog'),
-                 join_paths(meson.build_root(), 'README.html'),
-                 join_paths(meson.build_root(), 'ChangeLog.html'),
+                 join_paths(meson.source_root(), 'ChangeLog'),
+                 join_paths(meson.source_root(), 'README.html'),
+                 join_paths(meson.source_root(), 'ChangeLog.html'),
                  install_dir : helpdir)
     summary({'README' : ['README.html created:', true]}, section : 'Documentation', bool_yn : true)
 else
     install_data('README.md', 'COPYING', 'TODO', 'AUTHORS',
-                 join_paths(meson.build_root(), 'ChangeLog'),
-                 join_paths(meson.build_root(), 'ChangeLog.html'),
+                 join_paths(meson.source_root(), 'ChangeLog'),
+                 join_paths(meson.source_root(), 'ChangeLog.html'),
                  install_dir : helpdir)
     summary({'README' : ['pandoc not found - README.html created:', false]}, section : 'Documentation', bool_yn : true)
 endif


caclark:
I don't think this is the right way to go. Meson permits generated files to be completely separated from the sources - so, for example you could use:
`meson setup /tmp/build`
`ninja -C /tmp/build`

This change puts generated files back into the source directory.

I don't understand Meson, but...
Could you insert `set -x` at the start of both `gen_readme.sh` and `gen_changelog.sh`, then run:
`ninja -C build -v`

Then zip up and upload  `build/meson-logs/meson-log.txt`
Perhaps there might be a clue there.

gusnan:
Here you have the log:
[meson-log.tar.gz](https://github.com/BestImageViewer/geeqie/files/9269121/meson-log.tar.gz)

eli-schwartz:
I did say that, by the way.

> The exact status of that command should be listed in meson-logs/meson-log.txt

And sure enough, that command is exiting early due to 
```sh
[ ! -x "$(command -v git)" ] && exit 0
```

eli-schwartz:
I did say that, by the way.

> The exact status of that command should be listed in meson-logs/meson-log.txt

And sure enough, that command is exiting early due to 
```sh
[ ! -x "$(command -v git)" ] && exit 0
```

caclark:
Hello Andreas
Attached  is another small change to the configuration setup.  Perhaps you could run a check before I make a commit.
As Eli has said, after you made the chroot, the command `git` was not available, so the changelog files were not being regenerated (the version number generation would fail also). This patch should fix that.
[1017-1.diff.gz](https://github.com/BestImageViewer/geeqie/files/9275035/1017-1.diff.gz)

gusnan:
Yeah, that patch fixes it. Thanks guys!

PR #1018, opened 2022-08-01
Fix "geeqie --help" remote commands section translation

cpina:
The strings were marked to be translated in the:
"static RemoteCommandEntry remote_commands" as N_("some string")

Now it translates the strings at printing time.


cpina:
As far as I can see these two `printf_term` are the only ones that used `N_` instead of `_`:

caclark:
Fixed in commit 88b5fab823d

PR #1019, opened 2022-08-01
Correct catalan translation

cpina:
I didn't properly test it before because I was confused with the problem that this PR https://github.com/BestImageViewer/geeqie/pull/1018 fixes. Now it all looks good and I don't expect changes for a while.


caclark:
Fixed in commits 8449bfdad 6c783e87c

PR #1020, opened 2022-08-08
Renames variables to avoid conflicting with C++ keywords

xsdg:


caclark:
Fixed in commits 01dc3bc cb8a40c 424e5091

PR #1022, opened 2022-08-13
Fix appdata and icon installation

Hofer-Julian:
- Appdata used invalid tags
- Icon was not installed


caclark:
Implemented in commit eb9d0e103

PR #1027, opened 2022-08-31
Detect strverscmp support properly

AMDmi3:


caclark:
Implemented in commit 87042fa5

PR #1028, opened 2022-08-31
Fix index.html symlink

AMDmi3:
Create relative symlink, as absolute one will point to temporary build directory


caclark:
Implemented in commit c5d75369

PR #1029, opened 2022-08-31
Detect libtiff properly through pkgconfig

AMDmi3:


caclark:
Implemented in commit ab73ca79

PR #1036, opened 2022-09-27
Update ko.po

MarongHappy:
All newly translated.


caclark:
Implemented in commit 4905c95a

Issue #1006, opened 2022-07-24
We should switch Geeqie to a more modern coding style


xsdg:
Long story short, clang-format exists, and makes it fairly easy to adjust the coding style of an entire codebase in a way that doesn't actually break the code in question.
See here for a quick introduction: https://www.kernel.org/doc/html/latest/process/clang-format.html

[Geeqie's coding style](https://github.com/BestImageViewer/geeqie/blob/master/CODING.md#source-code-style) has always felt somewhat bizarre.  And especially specifying the use of tab characters that are shown as 4-characters-wide, it's _super easy_ for someone to, without realizing, write code that doesn't match the spec.  In general, there is no modern C/C++ coding style that requires tab characters, and we should follow that pattern -- indentation with spaces only.

More specifically, I feel like we should try to adopt/adapt an existing coding style, rather than to just create a brand new one that's unique for Geeqie.  As potential starting points, darktable has a clang-format configuration here:
https://github.com/darktable-org/darktable/blob/master/.clang-format

Also the clang-format website offers inherent support for using LLVM, Google, Chromium, Mozilla, Webkit, Microsoft, and GNU coding standards as a foundation: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#configurable-format-style-options

I am personally most familiar with Google style.  That said, it has benefits and drawbacks just like any other coding style.  That said, of the lot, the GNU standard feels incredibly antiquated compared to the other ones.  For instance, it states "1989 Standard C is widespread enough now that it is ok to use its features in programs."  That doesn't feel like the right fit for the general changes we're trying to make in Geeqie right now.

Obviously, if we do decide to switch Geeqie from C to C++, then it feels like the coding style change should happen during or after that transition.  But it seemed good to at least start the conversation now rather than later.

caclark:
This is about other project standards, not coding style...

I get a bit irritated about the mixed use of hyphens and underscores in file names. I would prefer the project to standardise on hyphens.

However, git loses the history if a file is simply renamed. This script:
https://gist.github.com/emiller/6769886
did not work for me on a one-file test (and took over 4 minutes).
But nevertheless, messing around with the git history would worry me.

Does anyone have a simple solution to file renaming while maintaining the git history?

mowgli:
Hi Colin,

Am So den 21. Aug 2022 um 17:56 schrieb Colin Clark:
[...]

PR #1038, opened 2022-10-01
Renames source files from using a mixture of hyphens and underscores to using hyphens everywhere

xsdg:
Addresses one of the points in #1006


xsdg:
First phase of #1039

Issue #1039, opened 2022-10-01
Switch Geeqie from C to C++


xsdg:
This has been discussed extensively on the mailing list, and after conversation between @caclark, @mowgli, and myself, it seems like this is the direction that the project is headed.

Related email threads:
* https://www.freelists.org/post/geeqie/The-future-of-Geeqie,6
* https://www.freelists.org/post/geeqie/gqpp-C-geeqie-experiment
* https://www.freelists.org/post/geeqie/Please-trial-C-geeqie

At this point, the prototypes I've built compile in both GCC (with `-fpermissive`, but no code changes) and clang (with significant code changes).  The staging for the project transition will go as follows:
1) Change from mix of hyphens and underscores in source filenames to only hyphens (see also #1006 )
2) Rename files from `*.c` to `*.cc`.  Also, add the changes to `meson.build` to get things compiling with GCC.  **This will temporarily break clang compilation**
3) Add the code changes required to get things compiling with clang as well.  Because there are thousands of code changes (mostly adding typecasts), this is a lot riskier than step 2, which is why it makes sense to perform it as a separate step.

caclark:
Hello Omari

I have made some changes to the code just to reduce the number of compiler warnings, but my ignorance of c++ is getting in the way.

As an example, in this code from around line 348 in search.cc:
`hard_coded_window_keys search_window_keys[] = {
	{GDK_CONTROL_MASK, 'C', N_("Copy")},` etc

Following lines could be changed from e.g.:
`0, GDK_KEY_Return, N_("View")},`  

which shows the error `.warning: invalid conversion from int to GdkModifierType'


to:
`GdkModifierType(0), GDK_KEY_Return, N_("View")},`
 or 
`(GdkModifierType)0, GDK_KEY_Return, N_("View")},`
or
`static_cast<GdkModifierType>(0), GDK_KEY_Return, N_("View")},`

none of which show a compile error.

Is there a correct way to eliminate the warning?

Thanks...

xsdg:
Either the 2nd or 3rd option.  If GdkModifierType were a C++-style enum class, the 3rd option would be slightly preferable.  But given that it's a C-style enum, it doesn't matter.

I _believe_ the first option is going to call some kind of constructor on the struct.  It's possible that this will set the provided int as the enum value.  It's possible that it does something else, but that passing `0` happens to do something reasonable.  Without knowing what the constructor does, it's impossible to understand what the `GdkModifierType(N)` construct means or does.

PR #1040, opened 2022-10-01
Switches Geeqie source files from being C to C++ (breaks compilation with clang)

xsdg:
Second phase of #1039

This doesn't make any meaningful code changes, and as such, should be fairly safe.

As noted, **this breaks compilation with clang**, which doesn't support `-fpermissive` for making C-style code compile as C++.  This will be fixed with the third phase of #1039, which will involve substantial code changes, and thus will likely warrant significantly more validation.


caclark:
OK, but cannot be loaded after #1038. By itself is ok.

Do you want this committed (do you still have write access to the repo?)?

I have been looking at deheader and cppcheck to eliminate unused includes and functions. That means small changes in lots of files. Does that affect your work if I commit them soon?

xsdg:
I can merge them both when I get home in ~9 hours

PR #1042, opened 2022-10-18
Allow to use xxdi, lightweight alternative to vim's "xxd -i"

voyageur:
xxdi [0] is a lightweight script that can replace "xxd -i", which is now needed to build geeqie. This is helpful on systems not running vim. This checks for its availability on the system and falls back to existing vim mode if not installed.

[0] https://github.com/gregkh/xxdi


caclark:
Implemented in commit ef70ba519

PR #1043, opened 2022-10-18
Add a doc option to build system

voyageur:
Currently, HTML and lua API documentations are built automatically if relevant program (yelp-build or doxygen) is found. This adds a doc option to control whether such documentation should be built and installed.

This helps for packaging and reproducible builds.


caclark:
Fixed via 4ae3567

PR #1045, opened 2022-10-19
Miscellaneous Meson cleanups to do more work during build

eli-schwartz:
Using run_command is a bit awkward and it turns out we don't need to, here.


caclark:
Implemented in e491df3 454314cc da7ddeb f7bc982 91d7a63a

PR #1046, opened 2022-10-20
meson: distribute changelog when running `meson dist`, and install it

### LINKED ISSUE #1044: Source releases should include ChangeLog files (and README.html)
When building from source releases with meson like current [2.0.1](https://github.com/BestImageViewer/geeqie/releases/download/v2.0.1/geeqie-2.0.1.tar.xz), it is not possible to have ChangeLog files.

They are not included in release tarball and the way to generate them in meson.build requires a git checkout. I worked around it in Gentoo ebuild by building and providing these files manually, but it would be great to have them in release tarball - and installed if found in the source directory (especially if gen_changelog.sh cannot be run - providing an existing version)

It would be great to have a similar change and tarball inclusion for README.html as pandoc has a lot of dependencies (for the scope of this file)

eli-schwartz:
Installing without a changelog is still supported, if building without git *and* without an official dist tarball.

Fixes #1044

### DIFF
--- a/gen_changelog.sh
+++ b/gen_changelog.sh
@@ -38,4 +38,14 @@ echo "</html>" >> "$builddir/ChangeLog.$$.new.html"
 rm "$builddir/ChangeLog.$$.old.html"
 mv -f "$builddir/ChangeLog.$$.new.html" "$builddir/ChangeLog.html"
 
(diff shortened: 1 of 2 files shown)


caclark:
Implemented in commit 4ae3567

PR #1049, opened 2022-11-14
Please don't hardcode lua version in lua package name

mattdm:
This fails on Fedora Linux, since our Lua 5.4 package is just called `lua`, not `lua53`. (It wouldn't work for our compat package for older versions either, because pkg-config for that is installed as e.g. `lua-5.1`.)

This _should_ work since `req_version` is still there checking the actual version.

If it is necessary to search for the package by this name for some distros, please at least make that a special-case check in addition to the "nameversion" one. Thanks!

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -368,12 +368,12 @@ lua_dep = []
 req_version = '>=5.3'
 option = get_option('lua')
 if not option.disabled()
-    lua_dep = dependency('lua5.3',  version : req_version, required : get_option('lua'))
+    lua_dep = dependency('lua',  version : req_version, required : get_option('lua'))
     if lua_dep.found()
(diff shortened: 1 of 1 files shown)


caclark:
Thanks.
Unfortunately this fails on the Xubuntu system I am using.
Is the information in the following link relevant to this?

https://github.com/mesonbuild/meson/issues/4553

mattdm:
> Is the information in the following link relevant to this?
> https://github.com/mesonbuild/meson/issues/4553

Yes, directly. Looks like the workaround in this comment https://github.com/mesonbuild/meson/issues/4553#issuecomment-442670449 should apply here -- it checks for several possible names.

PR #1051, opened 2022-11-21
Add meson option to allow disabling doxygen detection.

0-wiz-0:
This is needed for packaging systems that check the installed files but can't assume what is available on the build system.

### DIFF
--- a/doc/meson.build
+++ b/doc/meson.build
@@ -26,6 +26,8 @@ endif
 
 install_subdir(destdir, install_dir : helpdir, exclude_directories : 'lua-api/latex')
 
+option = get_option('doxygen')
+if not option.disabled()
 doxygen = find_program('doxygen', required : false)
 if doxygen.found() and running_from_git
     srcdir = join_paths(meson.source_root())
@@ -36,4 +38,5 @@ if doxygen.found() and running_from_git
     summary({'lua-api' : ['lua-api help file created:', true]}, section : 'Documentation', bool_yn : true)
 else
     summary({'lua-api' : ['doxygen not found - lua-api help file created:', false]}, section : 'Documentation', bool_yn : true)
+endif
 endif
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -25,6 +25,7 @@ option('gq_localedir', type : 'string', value : '', description : 'Location wher
 
 option('archive', type : 'feature', value : 'auto', description : 'archive files e.g. zip, gz')
 option('cms', type : 'feature', value : 'auto', description : 'color management system')
+option('doxygen', type : 'feature', value : 'auto', description : 'doxygen')
 option('djvu', type : 'feature', value : 'auto', description : 'djvu')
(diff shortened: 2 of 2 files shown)


gdt:
This is also needed for pandoc (I found out after trying the pkgsrc version that doesn't look for doxygen).

caclark:
I'm sorry, but unfortunately I have no understanding of packaging systems.
Does this problem also relate to every `find_program()` occurrence in all `meson.build` files?
e.g. `evince` etc.

0-wiz-0:
When doxygen or pandoc are found, more files are installed, so this is an obvious difference.

If finding evince causes the compiled program to have different behaviour, then it would be good to have an option to disable that too, yes.

caclark:
The following have a `find_program(  , required : false)` with no `disable` option in `meson_options.txt`:
`yelp-build` - Help files
`git` - lua-api help document and  changelog.html
`pandoc` - README.html
`evince`  - print preview

Should they all have an entry in the `meson_options.txt` file?

gdt:
It should be possible do disable them from the meson command line, at least the first 3, because they change the set of installed files.

As for evince, I would lean to yes, although the harm from trying to run evince later and failing is less serious, after one has transported a binary.  It is a failure of reproducible builds -- I am guessing geeqie is not at the point of dealing with that, but it is the way of the future, eventually.

So yes, they should all have options making them possible to disable.

(This seems like a meson bug; an option to find a program and behave differently must in general be disablable.)

eli-schwartz:
> (This seems like a meson bug; an option to find a program and behave differently must in general be disablable.)

This is clearly bogus. How is it a bug in the buildsystem framework if it offers the ability to find something in non-fatal mode? This is critical functionality in the toolbox of build definition handling functions -- build systems need to be able to do things like say "component X is mandatory, but depending on the system, there may be different implementations of it, so, first check for one implementation, and if that fails, check for the other, and if neither is found, error out".

Meson does of course provide rich functionality for project developers to describe their requirements:
- add an option "the-foobar-feature" of type "feature" to meson_options.txt
- *use* the option in lookups: `find_program('foobarprog', required: get_option('the-foobar-feature'))`

Meson will then report "Program foobarprog skipped: feature **the-foobar-feature** disabled" if you pass the disabled option. You can also select to make it required, or to make it check for and autodetect it.

Crucially, it remains the responsibility of the build files, not the framework, to determine which lookups are "supposed to" be optional, and mark them such. That much is a valid feature request against the `meson.build` files.

gdt:
It's entirely ok to look for something and use it.  What I meant is that it's a bug to offer facilties to do that without baking in an option for a caller of meson to tell it "don't look for this", because a a"find if present" with no disable option is pretty much guaranteed to cause packaging system problems.   That's what we encountered updating geeqie in pkgsrc.

gdt:
But, it doesn't matter if we agree on this -- thank you very much for adding the disable option!

Issue #1052, opened 2022-11-22
geeqie SIGBUS handler calls mmap which is async-signal safe, and in a way that must return EBADF


gdt:
### ISSUE TYPE
* Bug Report

### GEEQIE VERSION
```
Geeqie 2.0.1 GTK3
```

### OS / DISTRIBUTION

NetBSD and pkgsrc, but issue is clear from reading the sources.

### SUMMARY
geeqie calls mmap from a signal handler, `src/main.cc:sigbus_handler_cb`.   POSIX allows only async-signal-safe calls, and [mmap is not on the list](https://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html#tag_15_04_03).

Further, the handler calls `mmap` with `fildes` = `-1`.   [POSIX says that fildes has to be an open file](https://pubs.opengroup.org/onlinepubs/9699919799/functions/mmap.html) and that mmap must return `EBADF` if "The fildes argument is not a valid open file descriptor."

### WAY FORWARD

The code in geeqie is hard to understand.   It calls mmap to apparently unmap the page, assuming some beyond-standards mmap behavior.

It's really hard to tell from POSIX if it is valid to give a SIGBUS for an address that is within the range of a previous non-errored mmap call.  I more or less read it as not allowing that.

Overall, I lean to just removing this code.   Then, if there is a need, there could be a data structure that records faulting addresses, the handler could just update that, and code that reads from mmap could check that and treat that file as closed if a faulting address is reported.   And, with a test case that can reproduce this, the discussion could be less theoretical.

PR #1053, opened 2022-11-22
Disable SIGBUS handler

gdt:
This handler leads to undefined behavior.  Further, it calls mmap in a way that POSIX says must fail with EBADF.  This commit returns to simply taking SIGBUS if it occurs and exiting.  If that happens, then an issue can be created and a way forward discusssed.  See https://github.com/BestImageViewer/geeqie/issues/1052 for more context.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -1008,22 +1008,31 @@ void exit_program(void)
 	exit_program_final();
 }
 
-/* This code is supposed to handle situation when a file mmaped by image_loader
+/* This code attempts to handle situation when a file mmaped by image_loader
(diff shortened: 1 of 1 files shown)


gdt:
With this patch applied to 2.0.1 in a pkgsrc context, it builds and runs ok (when no other process is messing with image files).

caclark:
Thanks.
Should `exit(-1)` be `exit(EXIT_FAILURE)`?

gdt:
Yes, my pre-POSIX muscle memory is showing.  Will force-push that.

caclark:
Implemented in commit ffb0e76064e

Issue #1064, opened 2023-03-06
Approach to set marks/keywords based on files selected outside of Geeqie


xsdg:
This stems from https://www.freelists.org/post/geeqie/Any-good-way-to-set-marks-keywords-given-a-list-of-files

As part of my photography workflow, I will use a button on my camera to tag specific images, and then during my import workflow, I'll create a list (`good.txt`) containing the filenames of the tagged images.

My desired end state is for these images to have a certain keyword applied within geeqie.

I came up with two ideas for how to accomplish this, both requiring extensions to the remote command API
## Add a remote command feature to set a keyword on an image.

 This would allow me to accomplish the goal with the following steps:
 1) Open Geeqie in the relevant directory
 2) Use `geeqie -r --get-filelist:` to see the current list of files, in order.  Use these to calculate an index for each file in the directory
 3) Walk through the images in the directory using `geeqie --next`, and for each one that should have a tag, use the new feature to add a known keyword to that image.  Like `geeqie -r --keyword-add:tagged`

## Add a remote command feature to alter the image selection
This is pretty similar to the last approach, except instead of adding the keyword directly, it's just modifying the selection.  This could be done iteratively as above, with `--selection-add` or something like that.  Or it could be done by path, `--selection-add:<FILE>`j

(Either of these options should involve a reasonable complement of related commands.  So also `--selection-remove` or `--keyword-remove`, etc.)

caclark:
Looping through the file list and executing `geeqie --remote --list-add:<filepath>` will place the images in a new Collection.
[N.B. if filepath is not full or does not exist, Geeqie will crash]

From there it would seem reasonable to select the images in the collection,  right-click on Keywords and `Add keyword to all selected` - but that does not work....

At the moment I am working on changing the way collections are displayed - I am making a collection appear as a pseudo directory in the main window. When that is complete, keywords can be attached to images in the displayed collection pseudo-Dir as in any other directory.

 The options `--remote --get-selected` and variants would be worth implementing. At the moment all that is possible is to move the focus (and selection) to a single image, and the current selection is available only via any plugin.

xsdg:
That sounds good.  I'm thinking about using these as the new actions.  Let me know if I should change them, or if there's anything missing:

- `--get-selection` (returns currently-selected images in the same format as `--get-filelist:` — filename + class)
- `--selection-add:<FILE>` (so long as FILE is selectable; adds current image to selection if FILE is omitted)
[...]

PR #1068, opened 2023-03-13
Adds a `--get-selection` remote command

xsdg:
Adds a `--get-selection` remote command that shows selected files (including sidecars) in the same style as `--get-filelist:`

Part of #1064


xsdg:
(Going to use branches in this project rather than the forked project)

xsdg:
Looks like creating a branch in the original project conflicted with the auto-replication process from geeqie.org, so re-opening this

PR #1081, opened 2023-05-06
Drop stale intltool references

thesamesam:
Moved to gettext when porting to meson.


caclark:
Thanks. Incorporated in commit 9992c8eef7

thesamesam:
Thank you!

PR #1082, opened 2023-05-07
Don't use too new *_DEFAULT/*_NONE defines

gusnan:
To build on current Debian stable, I have to remove using these new glib G_TYPE_ and G_REGEX_ default/none-defines and instead use 0 cast to the proper type. This makes it build with older glib (Debian Bullseye have glib  2.66.8). The defines that are currently used are from Glib 2.74).

For example see here: https://docs.gtk.org/glib/flags.RegexCompileFlags.html

for the G_REGEX_DEFAULT.


caclark:
Implemented in commit 44e33e2745

PR #1087, opened 2023-05-15
Add missing #ifdef for DEBUG_FILEDATA

tomaszg7:
It looks like one ifdef is missing in DEBUG_FILEDATA code making it impossible to build Geeqie on my system.
```
../src/filedata.cc: In function ‘void file_data_dump()’:
../src/filedata.cc:707:34: error: ‘global_file_data_count’ was not declared in this scope
  707 |                 log_printf("%d", global_file_data_count);
      |                                  ^~~~~~~~~~~~~~~~~~~~~~
```

Simple fix adds ifdefs around body of `file_data_dump()` function which is used only for debugging anyway. Maybe more code should be ifdefed like `log_print_file_data_dump()` in `debug.cc`. Also not sure how/if DEBUG_FILEDATA was supposed to differ from DEBUG_FD. I didn't study the code that carefully ;)

Amends commit cd72fa8.

### DIFF
--- a/src/filedata.cc
+++ b/src/filedata.cc
@@ -697,6 +697,7 @@ FileData *file_data_ref(FileData *fd)
  */
 void file_data_dump()
 {
+#ifdef DEBUG_FILEDATA
 	FileData *fd;
 	GList *list;
 
@@ -716,6 +717,7 @@ void file_data_dump()
 
 		g_list_free(list);
 		}
+#endif
 }
 
 static void file_data_free(FileData *fd)


caclark:
DEBUG_FD() is a function to print the entire contents of the filedata hashlist as a filename and a reference count.
There is a problem in Geeqie that the filedata ref. counting is incorrect - the ref counts are not being decremented in all cases. This is not a problem for a small number of files, but someone using something like 100,000 files has noted a problem - I think that may be because memory is being used extensively by the filedatas not being released.

I intended to try to find the problem by scattering DEBUG_FD() calls throughout the code. But it is still just another bug to fix.

PR #1088, opened 2023-05-15
Replace deprecated g_memdup with g_memdup2

tomaszg7:
`g_memdup` is declared as deprecated in current GLib, safer replacement `g_memdup2` is available since GLib-2.68. It's probably fairly minor thing, but gets rid of one of countless build warnings.

Code is protected by GLIB_CHECK_VERSION macro, so should build on older distros as well (e.g. Debian 11 or Ubuntu 20.04 LTS).

See https://docs.gtk.org/glib/func.memdup2.html

### DIFF
--- a/src/exif.cc
+++ b/src/exif.cc
@@ -584,7 +584,11 @@ gchar *exif_item_get_data(ExifItem *item, guint *data_len)
 {
 	if (data_len)
 		*data_len = item->data_len;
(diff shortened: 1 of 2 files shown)


tomaszg7:
Right. I added another commit. Checked that there are no more cases g_memdups in the code now.

gusnan:
Thank you - much appreciated!

PR #1097, opened 2023-06-04
clang-tidy fixups

neheb:


caclark:
Thanks.
Implemented in commits 17bc28d6 1224a11f 45599b24 b84788ae 7a9627a4

tomaszg7:
I'm not sure about the commit with `auto`. In my opinion it decreases readability of code and brings no benefit. It makes sense to use it for various exotic types like iterators, lambda stuff and so on, but for simple pointer types it serves no purpose and makes it harder to spot errors.

neheb:
I disagree. Example:

```
auto a = (const char*)"";
a = "a";
const auto b = (const char*)"";
b = "b"; // compile error
```

auto avoids duplicating the type. The non pointer version also makes it clearer what is const and what is not.

tomaszg7:
Well, maybe I'm just used to looking at what the type is on the left hand side of equal sign :)

BTW. While at it, it might make sense to upgrade all casts from C-style to C++-style. Right now the code is a complete mixture.

neheb:
Yeah I'll do that in a different PR.

PR #1098, opened 2023-06-04
clang-tidy: C to C++ casts

neheb:


xsdg:
Could you include in the commit message the command that you ran to generate this commit, please?  (With the understanding that there were some tweaks after the fact)

neheb:
Done

neheb:
The undo conversion commit is suspicious. Anyone know if it's correct?

neheb:
What you're seeing is a difference between 32 and 64-bit builds. I tested on the former. I'll look into fixing this. The fix is probably just reinterpret_cast.

neheb:
You guys should really set up CI for PRs. On Linux, macos, and maybe freebsd if it builds there.

caclark:
Implemented in commits 5533498 d4f7d83 3ac742d 203b358

PR #1103, opened 2023-06-11
clang-tidy: use using

neheb:
Found with modernize-use-using


caclark:
Implemented in commit cc152b0

PR #1104, opened 2023-06-11
clang-tify: remove void

neheb:
Found with modernize-redundant-void-args


caclark:
Implemented in commit 696dbd53

PR #1109, opened 2023-06-16
clang-tidy: with more stuff enabled

neheb:


caclark:
Implemented in commits ade97e7 b9f3c35 cbca444

PR #1110, opened 2023-06-17
Add two missing headers for improving the build on NetBSD.

0-wiz-0:


caclark:
Thanks. Implemented in commits 217c92e0 01d8762d

PR #1113, opened 2023-06-18
Fix #ifdef for JPEG XL support

TobiX:
It's `HAVE_JPEGXL` everywhere else, but `HAVE_JXL` here...


caclark:
Thanks. Implemented in commit a147c36

PR #1114, opened 2023-06-18
Find and use libexecinfo (for backtrace()).

### LINKED ISSUE #1111: NetBSD: missing link against libexecinfo
When building geeqie-2.1 on NetBSD, linking fails with:
```
ld: src/geeqie.p/main.cc.o: undefined reference to symbol 'backtrace'
ld: /usr/lib/libexecinfo.so.0: error adding symbols: DSO missing from command line
```
On NetBSD, `backtrace()` is provided in libexecinfo, and since `main.cc` uses this (when the `execinfo.h` header is found, which NetBSD provides), it needs to link against `-lexecinfo` on NetBSD.

I don't speak meson well, can you please add this?

0-wiz-0:
Fixes build on NetBSD.

Closes #1111.


caclark:
Maybe with these changes it would run ok:

+ `libexecinfo_dep = []`
`option = get_option('execinfo')`


+ `# Include unconditionally dependency for NetBSD.`
`libexecinfo_dep = cc.find_library('execinfo', required : false)`

0-wiz-0:
Thanks, added that, please try again!

caclark:
Implemented in commits 0c3a475 4c3eb0a6

PR #1115, opened 2023-06-18
clang-tidy: more missing fixes

neheb:


caclark:
Implemented in commit 17c01cfa

PR #1118, opened 2023-06-20
clang-tidy: fix readability stuff

neheb:


caclark:
Implemented in commit 84312a7

PR #1119, opened 2023-06-21
Add ExifData as extra argument to exif_item_get_data_as_text()

kevinbackhouse:
I think this should fix the issue that was reported to Exiv2: https://github.com/Exiv2/exiv2/issues/2638
[`Metadatum::print`](https://github.com/Exiv2/exiv2/blob/931a40a746f5678dcc4625b06a2eb25fa4f00b34/src/metadatum.cpp#L10) needs to be given a pointer to the `ExifData`.


caclark:
Thanks for the work. Implemented in commit c45cca7

PR #1121, opened 2023-06-23
some cleanups

neheb:
Not sure about some of these.

### DIFF
--- a/src/bar-keywords.cc
+++ b/src/bar-keywords.cc
@@ -676,7 +676,7 @@ static void bar_pane_keywords_dnd_receive(GtkWidget *tree_view, GdkDragContext *
 		{
 		case TARGET_APP_KEYWORD_PATH:
 			{
-			auto path = (GList *)(gtk_selection_data_get_data(selection_data));
+			auto path = static_cast<GList *>(*reinterpret_cast<const gpointer *>(gtk_selection_data_get_data(selection_data)));
 			src_valid = keyword_tree_get_iter(keyword_tree, &src_kw_iter, path);
 			string_list_free(path);
 			break;
--- a/src/exiv2.cc
+++ b/src/exiv2.cc
@@ -819,7 +819,7 @@ gchar *exif_item_get_string(ExifItem *item, int idx)
 #else
 		std::string str = em->toString(); /**< @FIXME ?? */
 #endif
-		if (idx == 0 && str == "") str = em->toString();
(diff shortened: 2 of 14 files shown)


tomaszg7:
Not sure about "auto with g_new". In Geeqie it was customary in the code to declare variables at the beginning of the function body. To be consistent one should get rid of most of those declarations. Not sure it is worth the effort and how it would affect readability of the code.

neheb:
I meant the double pointer commit. Seems fishy.

caclark:
@neheb 
Hello Rosen
The double pointer change is a problem - the current code, after an earlier cast change, is a bug.
Unfortunately these cast changes confuse me...

This is the original code, which works:
`GList *path = keyword_tree_get_path(keyword_tree, &child_iter);`
`gtk_selection_data_set(selection_data, gtk_selection_data_get_target(selection_data), 8, (gpointer) &path, sizeof(path));`

The current code is:
`GList *path = keyword_tree_get_path(keyword_tree, &child_iter);`
`gtk_selection_data_set(selection_data, gtk_selection_data_get_target(selection_data), 8, reinterpret_cast<const guchar *>(&path), sizeof(path));`

which crashes when executed.

I do not see the correct way to recast this. Do you have an idea?
Thanks...

neheb:
Can't say I understand the code. The cast can be removed I think. The sizeof is operating on a pointer instead of the type.

If the parameter type is really a gpointer (void*). Cast can be removed or replaced with a const cast.

neheb:
Maybe https://github.com/BestImageViewer/geeqie/commit/b5b655ddd8efdad56498cde402c7aa0f0afc86ac is a hint.

neheb:
I just looked. I did not touch this code before.

caclark:
I'm sorry, I misinformed you. The problem is in line 679 of bar-keywords.cc:
`//~ auto path = (GList *)(gtk_selection_data_get_data(selection_data));`
`GList *path = *(gpointer *)gtk_selection_data_get_data(selection_data);`

If you would like, I can try to fix it.

neheb:
Just revert https://github.com/BestImageViewer/geeqie/commit/d4f7d83269fdb13acaa0e5bc25018a8bc75a0ee9

neheb:
Reverted here.

caclark:
Implemented in commits  ad80078 89f1d6cf 4eb87105 581e87f7

PR #1126, opened 2023-06-30
clang-tidy: add some missing frees

neheb:


neheb:
hrm oh well.

caclark:
> hrm oh well.

Well, actually, thanks. 
I now see that there are no tests of Geeqie's ability to display image files. I have been looking at that now - I guess the best that is possible is to set up a database of sample images of each of the formats that Geeqie claims to be able to handle, and include a meson test to check if they are displayable.

That will not be comprehensive because there are so many variations - but it may help.

PR #1129, opened 2023-07-05
run through include-what-you-use

neheb:


caclark:
@neheb 
Hello Rosen

There are some problems with this commit.
I think it is because there are missing #ifdef HAVE_ statements around some includes.

It will take me a while to fix that.

qarkai:
I was going to fix include-what-you-use warnings and came across this PR.
@neheb Have you applied include-what-you-use suggestions automatically or by hand?

neheb:
automatically. compilation failures I fixed manually.

Take over if you want.

Issue #1117, opened 2023-06-19
fails to compile with lua-5.4


oneforall2:
* Bug Report

### GEEQIE VERSION

``2.1`

```

<!-- What GTK toolkit is used to compile Geeqie? If you compile geeqie
yourself, please paste the config result at the end of configure. -->
```
gtk+-3.0 found: YES 3.24.38 (cached)
```




/usr/bin/ld: lua.cc:(.text+0x26): undefined reference to `luaL_checkudata(lua_State*, int, char const*)'

/usr/bin/ld: src/geeqie.p/lua.cc.o: in function `lua_image_get_path(lua_State*)':
lua.cc:(.text+0x14): undefined reference to `luaL_checktype(lua_State*, int, int)'
/usr/bin/ld: lua.cc:(.text+0x26): undefined reference to `luaL_checkudata(lua_State*, int, char const*)'
/usr/bin/ld: lua.cc:(.text+0x3e): undefined reference to `lua_pushstring(lua_State*, char const*)'
/usr/bin/ld: lua.cc:(.text+0x59): undefined reference to `lua_type(lua_State*, int)'
/usr/bin/ld: lua.cc:(.text+0x63): undefined reference to `lua_typename(lua_State*, int)'
/usr/bin/ld: lua.cc:(.text+0x7a): undefined reference to `lua_pushfstring(lua_State*, char const*, ...)'
/usr/bin/ld: lua.cc:(.text+0x8a): undefined reference to `luaL_argerror(lua_State*, int, char const*)'
/usr/bin/ld: src/geeqie.p/lua.cc.o: in function `lua_image_get_name(lua_State*)':
lua.cc:(.text+0xa4): undefined reference to `luaL_checktype(lua_State*, int, int)'
/usr/bin/ld: lua.cc:(.text+0xb6): undefined reference to `luaL_checkudata(lua_State*, int, char const*)'
/usr/bin/ld: lua.cc:(.text+0xce): undefined reference to `lua_pushstring(lua_State*, char const*)'
/usr/bin/ld: lua.cc:(.text+0xe9): undefined reference to `lua_type(lua_State*, int)'
/usr/bin/ld: lua.cc:(.text+0xf3): undefined reference to `lua_typename(lua_State*, int)'
/usr/bin/ld: lua.cc:(.text+0x10a): undefined reference to `lua_pushfstring(lua_State*, char const*, ...)'
[...]

PR #1130, opened 2023-07-07
include lua.hpp instead of .h

neheb:
It seems extern C is needed.

### DIFF
--- a/src/lua.cc
+++ b/src/lua.cc
@@ -24,9 +24,7 @@
 
 #define _XOPEN_SOURCE
 
-#include <lua.h>
-#include <lauxlib.h>
-#include <lualib.h>
+#include <lua.hpp>
 
 #include <stdio.h>
 #include <glib.h>


xsdg:
As mentioned in the bug, given that `lua.hpp` is not present in lua's github, I don't feel like we should rely on its presence.  It would be more reliable to manually use `extern "C"` for those three includes.

neheb:
lua.hpp is installed by Lua's Makefile.

xsdg:
It sounds like you have extra context on this.  It would help if you'd add that to the bug.  I searched for a good long while and couldn't figure out where `lua.hpp` comes from.

That said, we don't have control over how distros (particularly Gentoo) choose to build or package lua.  Gentoo's build system [specifically points at the lua's github](https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-lang/lua/metadata.xml).  People report that `lua.hpp` is present, but again, it's not clear how that happens, so it feels very tenuous to rely on.  `extern "C"` is easy and is guaranteed to work across distros.

neheb:
If Gentoo is the issue, probably better to rely on the Lua wrap instead.

neheb:
Another solution could be to upgrade to C++17 and use __has_include.

caclark:
I am really unclear about this pull request. Is the overall opinion that it should be committed?

neheb:
FWIW I found the error with https://github.com/neheb/geeqie/tree/cii

xsdg:
Howdy, Colin.  Sorry that I'm disappearing and reappearing so much.

My opinion is that we should go with your original solution of adding an explicit `extern "C" { … }` to the existing three lus includes in Geeqie's `lua.cc`.

I feel like using `lua.hpp` adds distro compatibility risk for minimal benefit.

neheb:
Updated.

caclark:
@neheb @xsdg 
Hello Omari, Rosen

I have no understanding of this subject. Unfortunately it is necessary for someone else to make a decision here about what to commit.

There is the pull suggested by Rosen, and there is the "extern C"  solution listed here #1117.

What is the way forward - one, the other, both, neither?

My only comment is that the best solution is the one that keeps people happy, rather than one that is merely technically correct.

Thanks for any suggestions.

neheb:
This PR does the extern C now.

caclark:
Thanks. Implemented in commit a132645e

PR #1132, opened 2023-07-08
various cleanups

neheb:


caclark:
Hello Rosen

Some of the patches here eliminate braces around single line blocks. This is a matter of coding style, rather than correctness.

My opinion is that the Geeqie coding style (in CODING.md) should *require* braces in these cases.

Other people will have a different opinion, of course.

One reason is that the only way I can debug something is by throwing in large numbers of debug print statements - I have  keystroke macro to do that. At times I have to insert braces in order to insert a debug statement, which is tedious.

There are other reasons that have been much discussed (in both views) on the Internet.

neheb:
One thing I want to do long term is to clang-format the whole codebase to get rid of various issues like mismatched indentation 

Unfortunately, I don't see a good way to do so. The codebase uses a weird style that I haven't been able to get clang-format to handle properly.

caclark:
To reformat small sections of code I use `astyle` with these options:
style=vtk
indent=force-tab
pad-oper
pad-header
unpad-paren
align-pointer=name
align-reference=name

https://astyle.sourceforge.net/astyle.html#_style=vtk

Not minimizing source code is something I feel strongly about. I think I will make a post on [Geeqie Freelists](https://www.freelists.org/archive/geeqie/) to see what other people think.

neheb:
removed for now. I'll fix it up later.

btw, clang-tidy has a check to add missing braces. but the style is problematic.

caclark:
Implemented in commits a10a17b, ae0b217

There is now a clang-tidy check in the project. This is intended just to put something in place.

neheb:
meson creates a clang-tidy target by default.

caclark:
@neheb 
Hello Rosen

I have found that there is a problem with the vector conversion change.
I have attached a log file created with `meson setup -Ddevel=enabled build` and a test file.

I have to change some things in the test setup because this problem is not detected.

[jxl-log.txt.gz](https://github.com/BestImageViewer/geeqie/files/12122886/jxl-log.txt.gz)
[IMG_20200308_194050.jxl.gz](https://github.com/BestImageViewer/geeqie/files/12122887/IMG_20200308_194050.jxl.gz)

src/image-load-jpegxl.cc
```diff
@@ -153,12 +154,9 @@ static uint8_t *JxlMemoryToPixels(const uint8_t *next_in, size_t size, size_t *s
...
 
   JxlDecoderDestroy(dec);
-  if (success){
-    return pixels;
-  } else {
-    free(pixels);
-    return nullptr;
-  }
+  if (success)
+    return pixels.data();
```
qarkai, line 158: Looks like dangling pointer.

qarkai: Fix in #1137.

PR #1134, opened 2023-07-20
Various fixes of GList handling

qarkai:
Mostly fixes are:
- Use `g_list_free_full` instead of `while`+`g_list_free` to avoid extra loop.
- Remove some wrappers, use `reinterpret_cast<GDestroyNotify>`.
- Use temporary variable as list iterator to save original list pointer.

### DIFF
--- a/src/bar-exif.cc
+++ b/src/bar-exif.cc
@@ -752,10 +752,11 @@ GList * bar_pane_exif_list()
 		ped = static_cast<PaneExifData *>(g_object_get_data(G_OBJECT(pane), "pane_data"));
 
 		list = gtk_container_get_children(GTK_CONTAINER(ped->vbox));
-		while (list)
+		GList *work = list;
+		while (work)
 			{
-			entry = static_cast<GtkWidget *>(list->data);
-			list = list->next;
+			entry = static_cast<GtkWidget *>(work->data);
+			work = work->next;
 			ee = static_cast<ExifEntry *>(g_object_get_data(G_OBJECT(entry), "entry_data"));
 			exif_list = g_list_append(exif_list, g_strdup(ee->title));
 			exif_list = g_list_append(exif_list, g_strdup(ee->key));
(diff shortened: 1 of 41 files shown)


caclark:
Implemented in commits  62f1509a de7a9c68 4065439 5955d3e9bda 542db871 47cc60dc 5f85850b 099f0bf0b1 5aa4c572

qarkai:
Out of curiosity: why not apply PR (merge or squash and merge) using GitHub interface? Copying commits looks like a lot of routine work.

caclark:
Because I have no idea what I am doing :-(
My relevant knowledge of computers is with 19 inch racks...

I have write permission for the main repository at geeqie.org, which is where my commits go (via the .patch file from the pull request). Any commit there is automatically synced with the GitHib repo - but Klaus has set that up, not me.

mowgli:
Hi,

well, there is many what can be done with git itself. Unfortunately,
Colin and I live not in the same region. Otherwise I would love to
transfer some knowledge about git.

The short way:
pull requests can be handled via git itself.

Imagine, you have setup the github remote under the name "github". Then
you can do a `git fetch github pull/<pull-id>/head`. The result is the
synthetic local ref named "FETCH_HEAD" which could be merged via `git
merge FETCH_HEAD`

You can even setup git to fetch pull requests per default but I
currently don't find the doc for that.

Regards
   Klaus
-- 
Klaus Ethgen                                       http://www.ethgen.ch/
pub  4096R/4E20AF1C 2011-05-16            Klaus Ethgen ***@***.***>
Fingerprint: 85D4 CA42 952C 949B 1753  62B3 79D0 B06F 4E20 AF1C

caclark:
I do not see this method of committing as a problem. It does not happen that often, and the time it takes me to make the commits  is trivial (compared to the time I spend looking at a monitor, trying to understand something)

I do not see this as a problem that needs to be solved.

PR #1135, opened 2023-07-22
Add CommandBuilder in order to drop goto in editor_command_parse

qarkai:


qarkai:
I guess Check-Build-Most-Options failed because libjxl is not installed:
```
Run-time dependency libjxl found: NO (tried pkgconfig and cmake)
...
5/53 Image_ IMG_20200308_194050.jxl                       FAIL 
```
Could you update test environment?

caclark:
Implemented in commit 908c0845

PR #1136, opened 2023-07-22
Fix some memory leaks

qarkai:
- Use vector to avoid memory leak on array reallocation
- Use unique_ptr to avoid memory leaks

### DIFF
--- a/src/bar-gps.cc
+++ b/src/bar-gps.cc
@@ -666,8 +666,6 @@ static void bar_pane_gps_write_config(GtkWidget *pane, GString *outstr, gint ind
 	gint zoom;
 	ChamplainMapSource *mapsource;
 	const gchar *map_id;
-	gchar *str = nullptr;
-	GString *buffer = g_string_new(str);
 	gdouble position;
 	gint int_position;
 	gint w, h;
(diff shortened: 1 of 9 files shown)


caclark:
@qarkai  @neheb @xsdg  @mowgli  @tomaszg7
Hello to the above, plus anyone else.

In this commit Arkadiy has shifted some variable declarations from the head of a function to the point of use.
In other pulls, Rosen has deleted braces around single line "if" statements.

These are a matter of format style, rather than correctness.

It looks like it is time to make a decision about which coding style Geeqie should adhere to in the future.

This pull request is not the right place, but I think the discussion/decision should be made somewhere.

There is a thread on https://www.freelists.org/archive/geeqie/

qarkai:
> Arkadiy has shifted some variable declarations from the head of a function to the point of use
> These are a matter of format style, rather than correctness.

Nope. These shifts allow early return from function without need to free allocated memory.
It's recommended by C++ Guidelines: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-introduce

caclark:
A great deal has been done to modernize the code recently by you and Rosen, maybe prompted initiated  Omari's change from C to C++.
My comment about defining a style for Geeqie still stands.
I do not think that people like me can be trusted to create code to a fixed style without the help of a code formatter.
Some .clang-format thing would be good. I will add something to the comment on freelists.

..... the GitHub workflow action from your latest pull has gone wild .....

PR #1137, opened 2023-07-22
Fix dangling pointer in jxl

qarkai:
Reformat code.


qarkai:
I guess this PR could be closed after 7ce9d97.

PR #1138, opened 2023-07-23
Remove UNUSED macro

qarkai:
C++ allows unnamed parameter.


caclark:
Implemented in commit 18a8f2e4

Issue #1091, opened 2023-05-17
Deprecated GTK stuff (WIP)


tomaszg7:
It's just a heads-up post. I got bored last afternoon after trying to compile Geeqie and decided to start to reduce the amount of deprecated GTK functions used in the code and cut down the number of warnings. I know next to nothing about GTK structure and wanted to learn a bit ;)

I pushed some commits to a branch, but a lot of things are not completed yet. 
https://github.com/tomaszg7/geeqie/tree/deprecated_gtk_stuff

I also hit some weird linking problem related to Lua, which I can't figure out (yet).

tomaszg7:
One of the big items would be getting rid of GtkStock stuff, replacing it with more modern way to use icons. I did partial work and replaced some icons according to guidance from 
https://developer-old.gnome.org/gtk2/stable/gtk2-Stock-Items.html and https://specifications.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html 
The new icons should be more theme-consistent. In my case it looks like this:
![gtk_icons](https://github.com/BestImageViewer/geeqie/assets/16195733/d5da2ff1-031b-4653-a4aa-cca27e24ade2)

There are some missing icons that can be added (here fullscreen for example), and not all of them are specified in freedesktop specification. It will allow also (maybe) to remove some of the inline icons from the code.

caclark:
> It will allow also (maybe) to remove some of the inline icons from the code.

Why would you like to remove some inline icons?

tomaszg7:
> Why would you like to remove some inline icons?

To replace them with standard icons present in already existing themes. For example, Geeqie now includes a trash icon
![gq-icon-trash](https://github.com/BestImageViewer/geeqie/assets/16195733/11997e46-34e1-47e0-80ad-b7db30a78bee)
which can safely be replaced with "edit-delete-shred" or "edit-delete".

Moreover, those icons do not adapt to the chosen icon theme, which might lead to inconsistent look or even readability problems.

And finally, as I understand, these icons would better be installed in /usr/share/icons/, not be inlined and bloat the executable. That's what I was able to understand from GTK docs and other desktop apps specifications. Then even if we do need to use some specific icons which are not generally available, themes might just include their versions of them or we could possibly supply versions for a few popular themes (e.g. hicolor, adwaita, breeze).

caclark:
Any new standard icons should be used, of course.

[...]

PR #1139, opened 2023-07-23
Remove some of deprecated GTK stuff

tomaszg7:
The aim of this branch was to get rid of some of deprecated code from old GTK versions. It was discussed in https://github.com/BestImageViewer/geeqie/issues/1091 It is not done, but with many changes to the codebase it would be better to have this bunch merged as it breaks quite easily with changes to Geeqie master branch. Particularly the discussion on coding style might lead to changes which will break these commits easily.

The commits here aim mostly on getting rid of GTK_STOCK. I managed to replace most of icons used in code with named icons, as current GTK coding style suggests. However there are some left throughout the code. There are also some problems as some icons are present only in particular icon themes, so a middle ground needs to be chosen.

To make further work easier, I tried to replace string literals for icon names with `#define` in `main.h`. Maybe with C++ work now under way it should be rather defined as `const string`, I don't know.

One big thing that still needs to be done is to remove the code which adds custom Geeqie icons to the stock and instead install them in `/usr/share/icons`. I started it but I need much more time to finish it, which I won't probably have in a near future.

I can't guarantee that there are no regressions, such as missing icons in some places, but as far as I could check, everything is in place. I need to do painful rebases on the way, but I think I didn't mess anything in the code.


src/print.cc
```diff
@@ -398,8 +398,8 @@ static void print_text_menu(GtkWidget *box, PrintWindow *pw)
 
 	hbox = pref_box_new(subgroup, FALSE, GTK_ORIENTATION_HORIZONTAL, PREF_PAD_BUTTON_GAP);
 
-	button = pref_button_new(nullptr, GTK_STOCK_SELECT_FONT, _("Font"), FALSE,
-				 G_CALLBACK(print_set_font_cb), const_cast<char *>("Image text font"));
+	button = pref_button_new(nullptr, GQ_ICON_SELECT_FONT, _("Font"),
+				 G_CALLBACK(print_set_font_cb), (gpointer)"Image text font");
```
qarkai, line 402: Probably should stay with c++ cast.
```suggestion
				 G_CALLBACK(print_set_font_cb), const_cast<char *>("Image text font"));
```

qarkai (commented):
Could you add named icons for `"document-properties"`, `"document-save-as"`, `"edit-delete-shred"`, `"edit-find-replace"`, `"go-jump"`?

tomaszg7:
@qarkai  thanks a lot for locating those missing parts.

src/img-view.cc
```diff
@@ -1594,13 +1594,13 @@ static GtkWidget *view_confirm_dir_list(ViewWindow *vw, GList *list)
 	g_signal_connect(G_OBJECT(menu), "destroy",
 			 G_CALLBACK(view_dir_list_destroy), d);
 
-	menu_item_add_stock(menu, _("Dropped list includes folders."), GTK_STOCK_DND_MULTIPLE, nullptr, nullptr);
+	menu_item_add_stock(menu, _("Dropped list includes folders."), GQ_ICON_DND, nullptr, nullptr);
```
qarkai, line 1597:
```suggestion
	menu_item_add_icon(menu, _("Dropped list includes folders."), GQ_ICON_DND, nullptr, nullptr);
```
or is it really have to be `menu_item_add_stock`?

tomaszg7: For now it needs to stay *_stock since I didn't find a correct replacement for `GTK_STOCK_DND_MULTIPLE` and `GQ_ICON_DND` resolves to that. Similar situation is with `GQ_ICON_COLLECTION`.

src/main.h
```diff
@@ -133,6 +133,89 @@
...
+#define GQ_ICON_ZOOM_OUT "zoom-out"
+#define GQ_ICON_ZOOM_100 "zoom-original"
+#define GQ_ICON_ZOOM_FIT "zoom-fit-best"
+
+// might need replacing
+#define GQ_ICON_PREFERENCES "preferences-system"
+#define GQ_ICON_HELP "help-contents" // "help-browser"?
+#define GQ_ICON_EXPORT "document-export" // use collection icon?
+
+// not available in some themes?
```
qarkai, line 192: Since "dialog-cancel" and "emblem-ok" are not listed in [Standard Icon Names](https://specifications.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html#names) there is little difference with "gtk-apply", "gtk-cancel" and "gtk-ok".
Gtk recommends replacing these icons with labels "_Apply", "_Cancel" and "_OK".

tomaszg7: I know, it is a problem with several stock icons not having a suggested replacement in standard icon names. I think that in most use cases the labels are already present (although sometimes they are different, like "Start"). I didn't have a good solution.

qarkai: What about following:
1) leave label if button already has it, remove icon
2) use suggested label if button doesn't have any
3) use some other icon if it applicable in context, e.g. "media-playback-start" icon for "Start" button

tomaszg7: I have no strong feelings on the matter. I just tried to preserve all the icons that were present as closely as possible. 

Generally I suppose if this bunch of commits gets merged, it will be possible to fine-tune the look. My preference would probably be also to remove some of the icons as I prefer cleaner look.

caclark:
@tomaszg7 
Hello Tomasz

Unfortunately this pull request now shows conflicts.

tomaszg7:
@caclark Ok, it should be resolved now.

caclark:
@tomaszg7 
Hello Tomasz
The file 1139.diff works. The file 1139.patch is totally screwed up. There must be a problem on GitHub.
The quickest solution would be for me to commit the .diff file, listing you as author.

tomaszg7:
Not sure exactly what those files you refer are, but feel free to solve the problem however suits you :) 

Like many people already told you, you seem to do things with Github the hard way. Simply clicking "merge" here and then pulling changes locally and pushing to official repo should do the trick.

qarkai:
> Not sure exactly what those files you refer are

https://github.com/BestImageViewer/geeqie/pull/1139.diff and https://github.com/BestImageViewer/geeqie/pull/1139.patch

> but feel free to solve the problem however suits you :)

@tomaszg7 Maybe if you squash commits then 1139.patch would work too.

> Like many people already told you, you seem to do things with Github the hard way. Simply clicking "merge" here and then pulling changes locally and pushing to official repo should do the trick.

IIUC GitHub repo is a mirror which is synced from [main source repository](http://geeqie.org/cgi-bin/gitweb.cgi?p=geeqie.git;a=summary), so probably 'GitHub way' is not an option.

tomaszg7:
> @tomaszg7 Maybe if you squash commits then 1139.patch would work too.
Thanks, I never tried to get these files out of github and didn't notice that the number is the number of PR.

I can try to squash commits, of course, but I feel that it should be the last resort as it's generally harder to read and backtrack changes.
 
> IIUC GitHub repo is a mirror which is synced from [main source repository](http://geeqie.org/cgi-bin/gitweb.cgi?p=geeqie.git;a=summary), so probably 'GitHub way' is not an option.

I know that. But if you push/commit/do something to this mirror, you should be able to use your local clone to pull from GH and then push back to the main repo. It will then try to sync itself with GH but that shouldn't do anything as the commits will already be there. As far as I understand git, it is really distributed and decentralized and no main from the point of view of git exists.

caclark:
Implemented in commit 3c328889

PR #1140, opened 2023-07-24
Fix usage of vectors

qarkai:
reserve() doesn't allocate elements.

### DIFF
--- a/src/color-man.cc
+++ b/src/color-man.cc
@@ -511,7 +511,7 @@ static cmsToneCurve* colorspaces_create_transfer(int32_t size, double (*fct)(dou
 		{
 		const double x = static_cast<float>(i) / (size - 1);
 		const double y = MIN(fct(x), 1.0f);
-		values[i] = static_cast<float>(y);
+		values.push_back(static_cast<float>(y));
 		}
 
 	return cmsBuildTabulatedToneCurveFloat(nullptr, size, values.data());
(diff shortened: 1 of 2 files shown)


src/color-man.cc
```diff
@@ -511,7 +511,7 @@ static cmsToneCurve* colorspaces_create_transfer(int32_t size, double (*fct)(dou
 		{
 		const double x = static_cast<float>(i) / (size - 1);
 		const double y = MIN(fct(x), 1.0f);
-		values[i] = static_cast<float>(y);
+		values.push_back(static_cast<float>(y));
```
tomaszg7, line 514: Wouldn't it make more sense to have iterator of `float` type now running over possible values of `x`? For example compute `1/(size-1)` and increment by that value?

tomaszg7: I was thinking about something like this:
```
const double delta = 1.0 / (size - 1);
for(double x = 0; x < 1; x += delta)
		{
		const double y = MIN(fct(x), 1.0f);
		values.push_back(static_cast<float>(y));
		}
```
(I didn't think through if the range is correct, needs a check)
The advantage would be that no float division is done inside the loop. Loop increment is float addition though but I still think it could be cheaper. I am not sure, I don't even know if the compiler would do this simplification on its own. Another caveat would be if small rounding errors wouldn't cause a problem.

caclark:
Implemented in commit 820c335

PR #1141, opened 2023-07-26
Fix memory leak in log_window_append()

qarkai:
Use deque and string.


caclark:
Implemented in commit e556ceaa

PR #1143, opened 2023-07-29
Remove deprecated GtkHPaned, GtkVPaned

qarkai:
Use recommended GtkPaned.


caclark:
Implemented in commit 090d786

PR #1144, opened 2023-07-29
Remove deprecated GTK_STOCK_DIALOG_*

qarkai:
Use standard icon names and rename related GQ_ICON macros.


caclark:
Implemented in commit c119aca

PR #1146, opened 2023-07-29
Remove deprecated GtkAlignment

qarkai:
Use GtkWidget:halign and GtkWidget:valign properties.


caclark:
Implemented in commit 3666d50

PR #1147, opened 2023-07-30
Replace MD5Context with GChecksum

qarkai:
- Replace md5_get_digest with md5_get_string
- Add md5_get_string_from_file
- Add MD5_SIZE constant


caclark:
Implemented in commit a4c408a8

PR #1152, opened 2023-07-31
Fix memory leaks in lua.cc

qarkai:


src/lua.cc
```diff
@@ -358,26 +359,19 @@ gchar *lua_callvalue(FileData *fd, const gchar *file, const gchar *function)
 {
 	gint result;
 	gchar *data = nullptr;
-	gchar *path;
 	FileData **image_data;
 	gchar *tmp;
```
xsdg, line 363: While you're here, maybe zero-initialize this pointer?

qarkai: I'd rather move declarations to their first usages. As in https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-introduce

xsdg: Because data and tmp are coupled, I think it makes sense to keep them defined (and declared) next to each other.

I would agree with moving the definition for a variable that's used in isolation, but in this case, I think it would increase the likelihood of follow-on bugs.

qarkai: To be honest I don't see how https://github.com/qarkai/geeqie/blob/lua-memleaks-2/src/lua.cc#L358 is more confusing than https://github.com/qarkai/geeqie/blob/lua-memleaks-1/src/lua.cc#L358

xsdg: Oh, yeah, that's fine.  I didn't realize that `data` also wasn't used until later.  As long as the data and tmp definitions are next to each other, I'm happy :+1:

xsdg: (to be clear, by "that" I mean `lua-memleaks-2`)

caclark:
Implemented in commit 43884274

PR #1153, opened 2023-08-02
Use pixbuf_fallback() instead of duplicated code

qarkai:


caclark:
Implemented in commit 96110e65d

PR #1154, opened 2023-08-03
Drop goto in copy_file()

qarkai:
Use lambda and smart pointers.


caclark:
Implemented in commit b8f365d

PR #1155, opened 2023-08-05
Drop goto in secure-save.cc

qarkai:
Use lambda and smart pointers.


caclark:
Implemented in commit 5d4aec8

PR #1157, opened 2023-08-07
Drop goto in zonedetect.cc

qarkai:
Use while loop and smart pointers.


caclark:
Implemented in commit 4c61dd2ee

PR #1161, opened 2023-08-11
Replace handmade spinner with GtkSpinner

qarkai:


caclark:
Implemented in commit 4ecbd04c5b02

[I guess it had to be, but... The old spinner had a 1960s sci-fi feel to it. Very retro. I am sad to see it go :-) ]

qarkai:
> The old spinner had a 1960s sci-fi feel to it. Very retro.

IIUC GtkSpinner uses `process-working` icon from current icon theme. Probably Geeqie could provide own theme overriding particular icons/animations and set it for itself.

caclark:
I think it best to leave it as is. Users are accustomed to whatever standard icons are in their installation. 
Going retro is maybe not a good idea - although some people think the whole of the Geeqie user interface is old-fashioned :-(

PR #1165, opened 2023-08-13
Remove some deprecated GdkScreen API

qarkai:


caclark:
Implemented in commit 101b77de

PR #1170, opened 2023-08-18
Remove some extra memory allocations

qarkai:
- Use `g_string_free(string, FALSE)`. Also use `len` member instead of `strlen(string->str)`.
- Use string literals instead of building string for resources' paths.

### DIFF
--- a/src/collect-io.cc
+++ b/src/collect-io.cc
@@ -198,7 +198,7 @@ static gboolean collection_load_private(CollectionData *cd, const gchar *path, C
 				}
 			}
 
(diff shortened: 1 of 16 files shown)


qarkai:
Not sure what to do with Geeqie AppImage build error `Error 403: Resource not accessible by integration`.

caclark:
Implemented in commit 23b57201

caclark:
> Not sure what to do with Geeqie AppImage build error Error 403: Resource not accessible by integration.

That is my mistake. Fixed in commit 5ab324f

PR #1172, opened 2023-08-20
Simplify GString usage

qarkai:
- Replace `g_string_append_printf` with `g_string_append` if format string doesn't contain format specifiers.
- Use `g_string_append_printf` instead of `g_string_append(string, g_strdup_printf)`.
- Merge `g_string_new` and following `g_string_append`.
- Remove unnecessary memory allocations and fix memleaks.


caclark:
Implemented in commit 495553c8ad

PR #1173, opened 2023-08-20
Simplify collection_save_or_append_dialog

qarkai:
- Use one comparison to detect what to do.
- Reduce scope of variables.
- Deduplicate code.
- Free GString memory.


caclark:
Implemented in commit f90512d1b2

PR #1174, opened 2023-08-21
Simplify gr_action_list

qarkai:
- Call `g_utf8_strlen` for `action_item->name` once.
- Remove redundant intermediate `list_final`.
- Reduce scope of `GList` iterators.


caclark:
Implemented in commit ee5b7a485d

PR #1175, opened 2023-08-23
Ensure actions listed are the shortest names

caclark:
The current implementation lists the ALT versions of some actions names.

### DIFF
--- a/src/ui-misc.cc
+++ b/src/ui-misc.cc
@@ -1427,8 +1427,10 @@ GList* get_action_items()
 	gchar *tooltip;
 	GList *actions;
 	GList *groups;
-	GList *list = nullptr;
-	GList *work;
+	GList *list_duplicates = nullptr;
+	GList *list_unique = nullptr;
+	GList *work1;
+	GList *work2;
 	GtkAction *action;
 	LayoutWindow *lw = nullptr;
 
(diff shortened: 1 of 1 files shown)


src/ui-misc.cc
```diff
@@ -1479,33 +1481,7 @@ GList* get_action_items()
 					action_item->name = g_strdup(action_name);
```
qarkai, line 1481: `action_name`, `label`, `tooltip` leak now.
```suggestion
					action_item->name = action_name;
```
Also for `action_item->label`.

src/ui-misc.cc
```diff
@@ -1514,9 +1490,40 @@ GList* get_action_items()
...
 
-	list = g_list_sort(list, simple_sort_cb);
+	/* Use the shortest name i.e. ignore -Alt versions. Sort makes the shortest first in the list */
+	list_duplicates = g_list_sort(list_duplicates, simple_sort_cb);
+
+	/* Ignore duplicate entries */
+	work1 = list_duplicates;
+	while (work1)
+		{
+		duplicate = FALSE;
```
qarkai, line 1500: Use `g_list_find_custom` instead?

caclark: Thanks

PR #1176, opened 2023-08-26
Use filelist_copy to reduce duplication

qarkai:
Simplify related code.

### DIFF
--- a/src/filedata.cc
+++ b/src/filedata.cc
@@ -1541,15 +1541,10 @@ void filelist_free(GList *list)
 GList *filelist_copy(GList *list)
 {
 	GList *new_list = nullptr;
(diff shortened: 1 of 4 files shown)


caclark:
Implemented in commit 9c740415fbf

caclark:
@qarkai 
Hello Arkadiy

I think there is a problem. If I right-click on an item in the files pane, there is a seg. fault.

[When I use git to revert the commit, the problem goes away. When I revert the commits by hand, the bug still exists....
Strange. But something is amiss somewhere.]

qarkai:
@caclark I messed with lists concatenation. I'l fix it now.

qarkai:
Fix in #1179.

PR #1177, opened 2023-08-26
Remove structs and some enums from typedefs.h

qarkai:
Move them to appropriate files.


caclark:
Implemented in commit dec9d67c4cdbf9

PR #1179, opened 2023-08-28
Fix GList concatenations

qarkai:
Introduced in #1176. `g_list_prepend` created heterogeneous lists.


caclark:
Implemented in commit fe565413b8e2

PR #1180, opened 2023-08-29
Move ClipboardDestination enum from typedefs.h to utilops.cc

qarkai:
Leftover from #1177.


caclark:
Implemented in commit 15d08126a5b02f

PR #1182, opened 2023-09-01
Do not truncate socket path

keszybz:
If the length of the socket filename is too long to fit in the socket structure, then we should fail, instead of possibly creating a socket with an unexpected name, possibly in the the wrong directory.

Also, in the structure, the filename is only followed by '\0' if the path is shorter than the maximum length. It is OK for the path use the full field, and it is unterminated in this case.

### DIFF
--- a/src/remote.cc
+++ b/src/remote.cc
@@ -246,9 +246,14 @@ static RemoteConnection *remote_server_open(const gchar *path)
 {
 	RemoteConnection *rc;
 	struct sockaddr_un addr;
-	gint sun_path_len;
 	gint fd;
 	GIOChannel *channel;
+
+	if (strlen(path) >= sizeof(addr.sun_path))
+		{
+		log_printf("Address is too long: %s\n", path);
+		return nullptr;
+		}
 
 	if (remote_server_exists(path))
 		{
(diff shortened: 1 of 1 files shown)


caclark:
Thanks for the contribution.

I have no experience of this kind of coding, so I apologize for querying this solution.

Is the event listed in the BUGS section of the sockets man page taken into account?

The `"Pathname sockets"` section of the man page states `"The length of the pathname, including the terminating null byte, should not exceed the size of sun_path."`

So maybe the test should be `if (strlen(path) + 1 > sizeof(addr.sun_path))` ?

keszybz:
I wrote the patch in a way that does not follow the recommendation in the man page. Instead, if allows the maximum possible length of the path (108 bytes). If the path is exactly 108 characters, then there is no terminating zero. The kernel is actually fine with that, it does not need the terminating zero.

Another option would be to rework the patch to limit the path to 107 bytes, which means that there's always space for the terminating zero. I can do that if you prefer.

caclark:
> Another option would be to rework the patch to limit the path to 107 bytes, which means that there's always space for the terminating zero. I can do that if you prefer.

Yes, I would prefer that approach.

Thanks again for the contribution.

keszybz:
Updated.

Thank you for maintaining geeqie.

caclark:
Implemented in commit ee970dfa957

PR #1191, opened 2023-09-24
Use hypot for Pythagorean addition

qarkai:


caclark:
Implemented in commit e036f9727a4f

PR #1193, opened 2023-09-30
Replace gtk_widget_render_icon with gtk_icon_theme_load_icon

qarkai:
* Use icon size in pixels in gtk_icon_theme_load_icon
* Deduplicate folder icons creation
* Copy pixbuf returned by gtk_icon_theme_load_icon to avoid GTK+ keeping the old icon theme loaded
* Move static functions to anonymous namespace
* Replace some icon name string literals with GQ_ICON macros

### DIFF
--- a/src/ui-misc.cc
+++ b/src/ui-misc.cc
@@ -1626,4 +1626,15 @@ gboolean defined_mouse_buttons(GtkWidget *, GdkEventButton *event, gpointer data
 	return ret;
 }
 
+GdkPixbuf *gq_gtk_icon_theme_load_icon_copy(GtkIconTheme *icon_theme, const gchar *icon_name, gint size, GtkIconLookupFlags flags)
+{
+	GError *error = nullptr;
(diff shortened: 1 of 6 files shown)


caclark:
Implemented in commit 51dfbeec

src/ui-misc.cc
```diff
@@ -1626,4 +1626,15 @@ gboolean defined_mouse_buttons(GtkWidget *, GdkEventButton *event, gpointer data
 	return ret;
 }
 
+GdkPixbuf *gq_gtk_icon_theme_load_icon_copy(GtkIconTheme *icon_theme, const gchar *icon_name, gint size, GtkIconLookupFlags flags)
+{
+	GError *error = nullptr;
+	GdkPixbuf *icon = gtk_icon_theme_load_icon(icon_theme, icon_name, size, flags, &error);
+	if (error) return nullptr;
```
caclark, line 1633: @qarkai 
Should there be a g_error_free() here at line 1633?

qarkai: You are right. Created #1429.

PR #1194, opened 2023-10-01
Replace deprecated *-gdk properties

qarkai:
Use *-rgba counterparts, rename related GQvCellRendererIcon properties


caclark:
Implemented in commit c6d324ab

PR #1195, opened 2023-10-01
Replace deprecated gtk_widget_get_state

qarkai:
Use gtk_widget_is_sensitive


caclark:
Implemented in commit 076e2c712c5

PR #1197, opened 2023-10-01
Use GdkRectangle for CollectionData::window

qarkai:


caclark:
Implemented in commit aac873927c

PR #1198, opened 2023-10-04
Use sigsetjmp instead of setjmp where appropriate

AMDmi3:
At least on FreeBSD, `setjmp`/`longjmp` and `sigsetjmp`/`siglongjmp` are different families of functions and cannot be mixed together, as these use different state structures (`jmp_buf` and `sigjmp_buf` respectively). Since `sigjmp_buf` is used in the Geeqie code, use corresponding `sigsetjmp` for it.


caclark:
Implemented in commit 8bc059f7d30

Issue #1125, opened 2023-06-28
Lua detection via pkg-config broken


alerque:
https://github.com/BestImageViewer/geeqie/blame/83725949ec8ca4948bd1cb21a9364a3d62976009/meson.build#L419

This is completely broken. It works if `pkg-config` returns something for "lua", but **not** for any of the other possibilities because Meson dies in the first iteration of the look because it is a hard requirement. It never even bothers looking for "lua5.3" because it errors out when not finding "lua".

caclark:
The way I think it works is - If `lua` is not found, then the other items in the list are checked until one is found.

For example, I modified meson.build to make a test:
`req_version = '>=5.4'`
`    foreach name : ['lua5.3', 'lua-5.3', 'lua53', 'lua54']`

and the result is:

`Dependency lua5.3 found: NO found 5.3.6 but need: '>=5.4'`
`Found CMake: /usr/bin/cmake (3.25.1)`
`Run-time dependency lua5.3 found: NO (tried pkgconfig and cmake)`
`Dependency lua-5.3 found: NO found 5.3.6 but need: '>=5.4'`
`Run-time dependency lua-5.3 found: NO (tried pkgconfig and cmake)`
`Dependency lua53 found: NO found 5.3.6 but need: '>=5.4'`
`Run-time dependency lua53 found: NO (tried pkgconfig and cmake)`
`Run-time dependency lua54 found: YES 5.4.4`

If I leave `lua` in the foreach statement, `lua5.4.4` is found immediately.

alerque:
I understand that's how you *expected* it to work, but I don't think it is. I'm not getting the fall through to lua5.3 and not getting the summary error on L#429 that I would expect to hit if it found no Lua at all. On the other hand if I either drop the first `'lua', ` from the array or just replace the whole array with `'lua5.3'` the rest happily works and finds my Lua (I'm packaging for Arch Linux here).

xsdg:
@alerque Just for the record, what version of meson are you building with?

xsdg:
Also, can you include a build log or any debugging build artifacts?  Here's what I see in my case (debugging something else, but it turned out to offer a data point for this as well):

```
[...]
Run-time dependency libraw found: NO (tried pkgconfig and cmake)
Run-time dependency lua found: NO (tried pkgconfig and cmake)
Run-time dependency lua5.3 found: YES 5.3.6
Program pandoc found: NO
[...]
```

alerque:
[...]

PR #1199, opened 2023-10-04
Fix iterating through lua dependency variants when lua is required

AMDmi3:
When lua is required and the build fails to find the first lua dependency variant from the list, build fails right away without trying other variants. Fix this by treating only the last variant as required. Also, move unversioned lua variant to the end to prefer versioned ones, which lessens chance to pick different version than expected 5.3 and provides more generic error message when lua is not found.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -411,8 +411,9 @@ lua_dep = []
 req_version = '>=5.3'
 option = get_option('lua')
 if not option.disabled()
-    foreach name : ['lua', 'lua5.3', 'lua-5.3', 'lua53']
-        lua_dep = dependency(name, version: req_version, required: get_option('lua'))
+    lua_names = ['lua5.3', 'lua-5.3', 'lua53', 'lua']
+    foreach name : lua_names
+        lua_dep = dependency(name, version: req_version, required: name == lua_names[-1] ? get_option('lua') : false)
         if lua_dep.found()
             break
         endif


caclark:
@AMDmi3 
This topic is also mentioned in #1125.

It seems that FreeBSD uses Meson 1.2.2, but I have difficulty interpreting some other distribution's information - e.g Debian stable shows meson as 1.0.1-5. Is that really so? 

What should be used as the least common denominator?

If so maybe the meson required version can be revised and the solution in #1125 used.

AMDmi3:
> It seems that FreeBSD uses Meson 1.2.2, but I have difficulty interpreting some other distribution's information - e.g Debian stable shows meson as 1.0.1-5. Is that really so?

See https://repology.org/project/meson/versions. Yes, Debian 12 has meson 1.0.1.

> What should be used as the least common denominator?

meson reference says that multiple names are supported in `dependency` since meson 0.60.0, so you may want to either use that and require meson >= 0.60.0 or use my solution which does not require newer meson version.

caclark:
@gusnan 

I assume that Debian Stable is the most conservative of the distributions that include Geeqie.
Do you have a suggestion which version of Meson should be set?

gusnan:
Debian stable has meson 1.0.1 - but if I understand #1125 correctly, 0.60.0 is enough for what you need? I would of course be fine with that requirement. I wouldn't be extremely much against if you want to increase the requirement above this though - Debian packaging is always done against unstable, and there we pretty much always have updated packages (In this case we have meson 1.2.1). 

If we need a backported geeqie and we increase the meson version requirement (I do personal geeqie backports), we'll of course need an updated meson for stable - there was a backport of meson for Debian 11 (by the debian meson maintainers), so I believe if one asks the meson maintainers for a backport for Debian 12, there's a big chance they'll do the same again. (I haven't looked at backporting the actual meson package, that might be simple too).

But, of course development (in the Debian world) should be primarily with unstable in mind, and there we are fine when it comes to requirements.

caclark:
Fixed in commit bc1d5eb67f0

Ref. #1125

PR #1203, opened 2023-10-27
Use GdkRectangle in ScreenData

qarkai:
Make ScreenData and related functions private.


caclark:
Implemented in commit 68c80ddab46a

PR #1205, opened 2023-10-28
Use C++ types for ScreenData list

qarkai:


caclark:
Implemented in commit 1196a1a13

PR #1206, opened 2023-10-28
Move fullscreen static functions to anonymous namespace

qarkai:
Also move internal defines to cc file.


caclark:
Implemented in commit d4eb6a02f009e

PR #1207, opened 2023-11-09
Update Slovak translation.

jose1711:


caclark:
Thanks. Implemented in commit bb8c73b8178

PR #1208, opened 2023-11-19
Simplify fullscreen_prefs_get_geometry()

qarkai:
* Use GdkRectangle as return value
* Convert pointers to references
* Add get_screen_default_geometry() to reduce number of warnings.


caclark:
Implemented in commit 37cd0e2cd67

PR #1210, opened 2023-11-20
Remove duplicated JPEG_MARKER macros

qarkai:


caclark:
Implemented in commit c420270bcfc6

PR #1211, opened 2023-11-20
Debug requires libdwarf, but this is just linked for C and not C++

gusnan:
I noticed that when trying to debug something, I didn't get any debug symbols, and discovered that this requires to link with the dwarf library, but this wasn't done - looking at meson.build, this was done, but only for C. See the patch, it should most probably be done for cpp now. I am far from an meson expert, but this patch works for me and enables debugging when I expect it.

But, as said, I am not a meson expert, so this might need more fixes.


caclark:
Implemented in commit 8f50b277ff1347

[
For information only - I use this script to install the files I need:
./scripts/devel-install.sh
]

Issue #1201, opened 2023-10-09
double free or corruption when viewing raw files from multiple makes and models


xiota:
### ISSUE TYPE
* Bug Report

### GEEQIE VERSION
```
Geeqie 2.1+git20231007-aedccfcb GTK3

Run-time dependency gtk+-3.0 found: YES 3.24.38
Run-time dependency glib-2.0 found: YES 2.78.0
```

### OS / DISTRIBUTION

Arch Linux

### SUMMARY

Possibly related to #1123

When opening or viewing a folder containing raw files (`cr2`, `nef`, `raf`), `geeqie` crashes/exits with the following message:
```
double free or corruption (out)
Geeqie fatal error
Signal: Abort
Code: Invalid permissions
Address: 0x000003e8000016c4
```

<details><summary>gdb backtrace</summary>

```
double free or corruption (out)

Thread 1 "geeqie" received signal SIGABRT, Aborted.
__pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at pthread_kill.c:44
44            return INTERNAL_SYSCALL_ERROR_P (ret) ? INTERNAL_SYSCALL_ERRNO (ret) : 0;
(gdb) bt
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at pthread_kill.c:44
#1  0x00007ffff5ab88a3 in __pthread_kill_internal (signo=6, threadid=<optimized out>) at pthread_kill.c:78
#2  0x00007ffff5a68668 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007ffff5a504b8 in __GI_abort () at abort.c:79
#4  0x00007ffff5a51390 in __libc_message (fmt=fmt@entry=0x7ffff5bc855d "%s\n") at ../sysdeps/posix/libc_fatal.c:150
#5  0x00007ffff5ac27b7 in malloc_printerr (str=str@entry=0x7ffff5bcb3c8 "double free or corruption (out)") at malloc.c:5765
#6  0x00007ffff5ac4830 in _int_free_merge_chunk (av=av@entry=0x7ffff5c02ac0 <main_arena>, p=0x5555561d3b80, size=980096) at malloc.c:4669
[...]

PR #1214, opened 2023-12-04
exiv2: Fix use-after-free and double-free with exiv2 0.28.x

sgn:
In exiv2 0.28.x, DataBuf is only a wrapper for std::vector. Exiv2::DataBuf::data() returns a shallow copy of its underlying data.  On Exiv2::DataBuf::reset(), all of its data is invalidated, includes the data returned by Exiv2::DataBuf::data(). Thus, all usages of said data is use-after-free, attempt to use its later is double-free.

Make a copy ourselves instead.  This method works for the old exiv2, too.
Since Exiv2::PreviewImage::size() is available from forever, and both
Exiv2::PreviewImage::copy() and Exiv2::PreviewImage::pData() are
introduced in a same change (commit https://github.com/Exiv2/exiv2/commit/9578e8123 in 2008).

Fixes: #1201
Fixes: https://github.com/void-linux/void-packages/issues/47547

### DIFF
--- a/src/exiv2.cc
+++ b/src/exiv2.cc
@@ -1219,19 +1219,13 @@ guchar *exif_get_preview(ExifData *exif, guint *data_len, gint requested_width, 
 
 			Exiv2::PreviewImage image = pm.getPreviewImage(*pos);
 
-			Exiv2::DataBuf buf = image.copy();
-
-#if EXIV2_TEST_VERSION(0,28,0)
-                       *data_len = buf.size();
-                       auto b = buf.data();
-                       buf.reset();
-                       return b;
-#else
-			std::pair<Exiv2::byte*, long> p = buf.release();
-
-			*data_len = p.second;
-			return p.first;
-#endif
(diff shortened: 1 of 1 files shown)


xiota:
Here's a patch that seems to work: 
[exiv2.patch.gz](https://github.com/BestImageViewer/geeqie/files/13552416/exiv2.patch.gz)

sgn:
Both patches are doing the same thing.

```
Address: 0x0
```

Did you use `-fno-exceptions` and your machine is out-of-memory when you try my patch.

xiota:
My computer is not out of memory.  CXXFLAGS are default for Arch.  I'll test your revision mementarily.

xiota:
Would you stop force pushing changes?  I have no idea which version of the patch is being tested when you do that.

xiota:
f613468dd3189077fa4f09a8071c6554b41a47a6 works as expected.

caclark:
Implemented in commit 133147d4

PR #1219, opened 2023-12-16
Add loongarch64 support

gusnan:
I've got a bugreport on debian requesting loongarch64 architecture support for geeqie, with this patch included. Please merge it  - thanks in advance.

Debian bug here: https://bugs.debian.org/1058719

### DIFF
--- a/src/backward.h
+++ b/src/backward.h
@@ -4250,6 +4250,8 @@ public:
     #else
       error_addr = reinterpret_cast<void *>(uctx->uc_mcontext.pc);
     #endif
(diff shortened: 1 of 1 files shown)


caclark:
Implemented in commit decb47a650

212dandan:
The Debian bugid, please see: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1058719
Thanks for your help to submit this patch (geeqie-add-support-for-loongarch64.patch).

PR #1225, opened 2024-01-06
Fix meson finding webp_pixbuf_loader

gusnan:
Without this fix meson setup fails on Debian with

```meson.build:557:26: ERROR: Array index 1 is out of bounds for array of size 1.```


caclark:
Fixed in commit b765d90a

PR #1229, opened 2024-01-25
Add keywords to desktop file

gusnan:
Add Keywords to the desktop file - I've got a warning of the debian package analyser tool lintian, this commit fixes this, full lintian output here:

```
I: geeqie: desktop-entry-lacks-keywords-entry [usr/share/applications/org.geeqie.Geeqie.desktop]
N: 
N:   This .desktop file is either missing a Keywords entry, or it does not contain keywords above and beyond those already present in the Name or
N:   GenericName entries.
N:   
N:   The Keywords field is intended to show keywords relevant for a .desktop file.
N:   
N:   Desktop files are organized in key-value pairs and are similar to INI files.
N:   
N:   The desktop-file-validate tool in the desktop-file-utils package may be useful when checking the syntax of desktop entries.
N: 
N:   Please refer to https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s06.html, Bug#693918, and
N:   https://wiki.gnome.org/Initiatives/GnomeGoals/DesktopFileKeywords for details.
N: 
N:   Visibility: info
N:   Show-Always: no
N:   Check: menu-format
N: 
```


caclark:
Implemented in commit b5975d50b55eb

PR #1234, opened 2024-01-27
Use GdkRectangle for LayoutOptions::log_window

qarkai:
Reset log_window width/height values after upgrade since option names change from `"log_window.w"`/`"log_window.h"` to `"log_window.width"`/`"log_window.height"`.


caclark:
Implemented iin commit e19e7249b2

PR #1243, opened 2024-02-08
Include backward.h where it is actually used

qarkai:


caclark:
Implemented in commit d85af0f4b3d2da

PR #1245, opened 2024-02-10
Cleanup main.h header

qarkai:
* Include `compat.h`, `config.h`, `debug.h`, `intl.h`, `options.h`, `typedefs.h` where they are actually used.
* Remove standard library includes.
* Remove unused gdk-pixbuf includes.
* Make other headers self-contained.
* Move defines not related to `main.cc` from `main.h` to `main-defines.h`.


caclark:
What do you think about ordering the includes alphanumerically?
The document: https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes suggests this should be done.

There is little work to achieve this for this pull request:
Apply your pull request
Apply the attached diff
Run `./scripts/clang-tidy-check.sh --fix`

If this is done .clang-tidy in the repository can be amended so that all future pull requests have to comply.

[clang-tidy-1.diff.gz](https://github.com/BestImageViewer/geeqie/files/14238000/clang-tidy-1.diff.gz)

qarkai:
I believe it could be done in separate PR since it has another purpose. Also this PR is already huge.
I was going to add fixes of include-what-you-use warnings in this PR but changed my mind due to patch size.
So I'd create PR with proposed patch after the current one is merged.

qarkai:
Tried clang-tidy patch.
> Apply your pull request

Done.
> Apply the attached diff

Done.
> Run `./scripts/clang-tidy-check.sh --fix`

Done, nothing changed. Is it expected?

qarkai:
I see now. `Apply your pull request` mean `change files so that the script finds them'.

`./scripts/clang-tidy-check.sh --all --fix` changed another 200 lines (100 added, 100 removed).
I still prefer to sort headers in separate PR.

caclark:
Implemented in commit ca2b7be34cb

Issue #1244, opened 2024-02-09
Code style


qarkai:
1. `Variables declarations should be followed by a blank line and should always be at the start of the block` in https://github.com/BestImageViewer/geeqie/blob/master/CODING.md#source-code-style contradicts rule [ES.21](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es21-dont-introduce-a-variable-or-constant-before-you-need-to-use-it): `Don’t introduce a variable (or constant) before you need to use it`. I suggest removing this line from `CODING.md`.
2. I'm working on headers cleanup now (making them self-contained, fixing include-what-you-use warnings) but couldn't find rule for order of includes. Is there any? It no I suggest to pick one, e.g. https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes and describe it in `CODING.md`.

caclark:
> I'm working on headers cleanup now (making them self-contained, fixing include-what-you-use warnings) but couldn't find rule for order of includes. Is there any? It no I suggest to pick one, e.g. https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes and describe it in CODING.md.

I do not see any other definition. A reference to that link could be put in CODING.md.

The .clang-tidy file could be amended to include a check for this:
` +  llvm-*,`
` +  -llvm-qualified-auto,`
Unfortunately clang-tidy FIXIT only sorts alphanumerically and not the order of the types of headers.

It looks like a lot of work.

> Variables declarations should be followed by a blank line and should always be at the start of the block in https://github.com/BestImageViewer/geeqie/blob/master/CODING.md#source-code-style contradicts rule [ES.21](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es21-dont-introduce-a-variable-or-constant-before-you-need-to-use-it): Don’t introduce a variable (or constant) before you need to use it. I suggest removing this line from CODING.md.

Ah..... This is something else. I will think some more about this.

PR #1246, opened 2024-02-13
Sort headers using clang-tidy

qarkai:
Unfortunately clang-tidy don't sort blocks, only includes inside blocks.
Added include order rule to `CODING.md`.
Patch for clang-tidy is from https://github.com/BestImageViewer/geeqie/pull/1245#issuecomment-1938546860.
Related to #1244.


caclark:
Implemented in commit b31104586cfe5

PR #1249, opened 2024-02-16
Remove redundant GdkRGBA memcpy

qarkai:


caclark:
Implemented in commit 6229ff1e4

PR #1251, opened 2024-02-18
Bump Exiv2 minimum version requirement to 0.18

qarkai:
Remove obsolete code.

According to [repology](https://repology.org/project/exiv2/versions) 0.18 is the oldest version in repos.


caclark:
Implemented in commit 68417ca0fa74ac

PR #1252, opened 2024-02-18
Fix include-what-you-use warnings

qarkai:
Use `#if` to check 0/1 config value.

Skipped `whereami.cc`. Probably it should be moved to another directory, something like `thirdparty` or even use as git submodule.

### DIFF
--- a/geeqie.imp
+++ b/geeqie.imp
@@ -1,0 +1,32 @@
+[
+    { include: [ "<bits/types/sig_atomic_t.h>",        private, "<csignal>",               public ] },
+
+    { include: [ "\"cairo.h\"",                        private, "<cairo.h>",               public ] },
+
+    { include: [ "@\"champlain/champlain-[a-z-]+.h\"", private, "<champlain/champlain.h>", public ] },
+
+    { include: [ "@<exiv2/[a-z_]+.hpp>",               private, "<exiv2/exiv2.hpp>",       public ] },
+    { include: [ "<exiv2/exv_conf.h>",                 private, "<exiv2/exiv2.hpp>",       public ] },
+    { include: [ "<exiv2/xmp_exiv2.hpp>",              private, "<exiv2/exiv2.hpp>",       public ] },
+
+    { include: [ "\"gdk/gdkkeysyms.h\"",               private, "<gdk/gdk.h>",             public ] },
+
+    { include: [ "@\"gobject/.*\"",                    private, "<glib-object.h>",         public ] },
+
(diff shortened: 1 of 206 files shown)


qarkai:
> I looked at a small number of source files. `iwyu` has added includes that are not necessary e.g. `ui-help.cc` `#include <gdk/gdk.h>` `#include <glib-object.h>`
> 
> Are these defaults from `iwyu`?

`iwyu` doesn't allow transitive includes. So it suggests to add `gdk.h` for Gdk symbols, `glib-object.h` for GObject symbols even if they included in `gtk.h` in `ui-help.h`. I guess includes in h-file are not transitive for respective cc-file.

> `backward.h`, `zonedetect.cc` and `zondetect.h` are also verbatim copies (although things may have changed since then).

I recall making changes in zonedetect. Didn't know it was external.

> Putting all these files in a `thirdparty` directory (or whatever is the standard naming convention) means they can be eliminated from the checks done during a pull request.

That would be great. Not in this PR though.

caclark:
Implemented in commit c351a3a4fc69

caclark:
@qarkai 
For information only.

I started to create a GitHub action to compile-check Geeqie in a FreeBSD vm, but found some problems, some maybe originating in this pull request:

`./pan-view/pan-util.cc:24: //~ #include <features.h>`

`features.h` does not seem to be in FreeBSD

------

`../src/misc.cc:285:29: error: use of undeclared identifier 'LC_ALL'`
`current_locale = setlocale(LC_ALL, NULL);`

`#include <local.h>`
needs to be in `main.cc`, `misc.cc`, `metadata.cc` (maybe).

-------

`../src/layout-util.cc:611:16: error: use of undeclared identifier 'WEXITSTATUS'`
`run_result = WEXITSTATUS(cmdstatus);`

`#include <sys/wait.h>`
needs to be in layout-util.cc

-----

I have no idea about this stuff. I will wait for a FreeBSD packager to complain. Maybe they have ideas about how to solve the problems.
Or, if they do not complain, there is no problem (except that I cannot run the GitHub action without a fail being generated).

qarkai:
> `#include <features.h>`

According to glibc dev `features.h` is a [private header](https://stackoverflow.com/a/71058721/6771534). I'll remove it and fix iwyu rules in `geeqie.imp`.

> `undeclared identifier 'WEXITSTATUS'`

Apparently on my machine and Linux in general `WEXITSTATUS` is defined in `cstdlib`. I'll take a look.

PR #1253, opened 2024-02-19
Deduplicate cr3 image loader

qarkai:
Make it wrapper of jpeg image loader.
Add `HAVE_RAW` define check to make consistent with usage in `image-load`.
Move static functions to anonymous namespace.


caclark:
Implemented in commit 9336f097b6

PR #1256, opened 2024-02-20
Fix HAVE_* config values checks

qarkai:
c351a3a4fc698924a49dca09991aa0c5192ec345 follow-up.


caclark:
Implemented in commit 5e55a870557

PR #1257, opened 2024-02-21
Mark another string in the --help output as translatable

gusnan:
Just marking another string that should be translatable as such.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -551,7 +551,7 @@ static void parse_command_line(gint argc, gchar *argv[])
 #endif
 
 				print_term(FALSE, "\n");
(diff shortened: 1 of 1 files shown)


caclark:
Thanks @gusnan 

I have recently done some work to update the `untranslated-text.sh` script. The problem is to find strings without too many false hits or real misses.

The script does not detect this string in main.cc because it does not start with an alpha....

I will make all the missed translation updates in one go, at some time.

For information, I have included a .diff of the work so far.

Thanks again.
[untranslated-1.diff.gz](https://github.com/BestImageViewer/geeqie/files/14358517/untranslated-1.diff.gz)

PR #1259, opened 2024-02-24
Fix deprecation warning for poppler >= 0.82

qarkai:


caclark:
Implemented in commit db0c7d5

PR #1262, opened 2024-02-27
Simplify usage of set_page_num and get_page_total

qarkai:


caclark:
Implemented in commit 971800aeb852b3d

PR #1264, opened 2024-02-29
For POTFILES, look for whereami and zonedetect in third-party folder

gusnan:
Without this `meson compile geeqie-pot` won't work.


caclark:
Implemented in commit 0e1d21a7de86c

PR #1266, opened 2024-03-02
Move open_archive() to separate module

qarkai:
Also make parameter const and move static functions to anonymous namespace.


caclark:
Implemented in commit d27792e1ee9

PR #1267, opened 2024-03-03
Remove set but not used variables

qarkai:


caclark:
Implemented in commit 2c9ae87090a76

PR #1268, opened 2024-03-03
Compile modules based on HAVE_* features

qarkai:
Allows to reduce build time, `<config.h>` includes and number of `#if` macros a bit.

### DIFF
--- a/src/bar-gps.cc
+++ b/src/bar-gps.cc
@@ -19,11 +19,6 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
-#include <config.h>
-
-#if HAVE_LIBCHAMPLAIN
-#if HAVE_LIBCHAMPLAIN_GTK
-
 #include "bar-gps.h"
 
 #include <cstring>
@@ -37,6 +32,8 @@
 #include <gdk/gdk.h>
 #include <glib-object.h>
 #include <pango/pango.h>
+
+#include <config.h>
 
 #include "bar.h"
 #include "compat.h"
@@ -138,7 +135,7 @@ static void bar_pane_gps_close_save_cb(GenericDialog *, gpointer data)
 	g_list_free(pgd->geocode_list);
 }
 
- static void bar_pane_gps_dnd_receive(GtkWidget *pane, GdkDragContext *,
+static void bar_pane_gps_dnd_receive(GtkWidget *pane, GdkDragContext *,
 									  gint x, gint y,
 									  GtkSelectionData *selection_data, guint info,
 									  guint, gpointer)
@@ -1158,8 +1155,5 @@ void bar_pane_gps_update_from_config(GtkWidget *pane, const gchar **attribute_na
 	gtk_widget_set_size_request(pgd->widget, -1, pgd->height);
 	bar_update_expander(pane);
 }
-
-#endif // HAVE_LIBCHAMPLAIN_GTK
-#endif // HAVE_LIBCHAMPLAIN
 
 /* vim: set shiftwidth=8 softtabstop=0 cindent cinoptions={1s: */
--- a/src/bar.cc
+++ b/src/bar.cc
@@ -188,8 +188,7 @@ static const gchar default_config_copyright[] =
 "    </layout>"
 "</gq>";
 
-#if HAVE_LIBCHAMPLAIN
-#if HAVE_LIBCHAMPLAIN_GTK
+#if HAVE_LIBCHAMPLAIN && HAVE_LIBCHAMPLAIN_GTK
 static const gchar default_config_gps[] =
 "<gq>"
 "    <layout id = '_current_'>"
@@ -202,7 +201,6 @@ static const gchar default_config_gps[] =
 "        </bar>"
 "    </layout>"
 "</gq>";
-#endif
 #endif
 
 static const KnownPanes known_panes[] = {
@@ -218,10 +216,8 @@ static const KnownPanes known_panes[] = {
 	{PANE_EXIF,		"file_info",	N_("File info"),	default_config_file_info},
 	{PANE_EXIF,		"location",	N_("Location and GPS"),	default_config_location},
 	{PANE_EXIF,		"copyright",	N_("Copyright"),	default_config_copyright},
-#if HAVE_LIBCHAMPLAIN
-#if HAVE_LIBCHAMPLAIN_GTK
+#if HAVE_LIBCHAMPLAIN && HAVE_LIBCHAMPLAIN_GTK
 	{PANE_GPS,		"gps",	N_("GPS Map"),	default_config_gps},
-#endif
 #endif
 	{PANE_UNDEF,		nullptr,		nullptr,			nullptr}
 };
--- a/src/exif.cc
+++ b/src/exif.cc
@@ -45,30 +45,28 @@
  *     Add support for marker tag 0x0000
  */
 
-#include <config.h>
+#include "exif.h"
 
-#if !HAVE_EXIV2
-
-#include <cmath>
-#include <cstdio>
-#include <cstring>
 #include <fcntl.h>
 #include <sys/mman.h>
 #include <sys/stat.h>
-#include <sys/types.h>
 #include <unistd.h>
+
+#include <cstdio>
+#include <cstring>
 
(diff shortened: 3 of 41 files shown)


caclark:
Running `test-all.sh` gives me these errors:

test:         Code Correctness_ bar-gps.cc
start time:   10:32:34
duration:     7.90s
result:       exit status 1
command:      MALLOC_PERTURB_=187 /usr/bin/clang-tidy -p ./build -quiet --config-file /home/cclark/data/geeqie/geeqie/.clang-tidy /home/cclark/data/geeqie/geeqie/build/../src/bar-gps.cc
----------------------------------- stdout -----------------------------------
../src/bar-gps.cc:57:1: error: replace macro with enum [modernize-macro-to-enum,-warnings-as-errors]
#define THUMB_SIZE 100
^~~~~~~~
../src/bar-gps.cc:57:9: error: macro 'THUMB_SIZE' defines an integral constant; prefer an enum instead [modernize-macro-to-enum,-warnings-as-errors]
#define THUMB_SIZE 100
        ^
../src/bar-gps.cc:59:1: error: replace macro with enum [modernize-macro-to-enum,-warnings-as-errors]
#define DIRECTION_SIZE 300
^~~~~~~~
../src/bar-gps.cc:59:9: error: macro 'DIRECTION_SIZE' defines an integral constant; prefer an enum instead [modernize-macro-to-enum,-warnings-as-errors]
#define DIRECTION_SIZE 300

In commit a4ce95bc11a I made some changes for `modernize-macro-to-enum` and removed the exclude from `.clang-tidy`. I missed `bar-gps.cc`, but I do not know why.

The inconsistency in checks is because the GitHub runner is Ubuntu 22.04 which uses clang-tidy 14. The `modernize-macro-to-enum` check was introduced in clang-tidy 15.

At sometime the GitHub runner will be updated, but I do not know the schedule.

Do you wish the patch to be committed as it is?

qarkai:
Let's commit as it is. I'll fix these warnings in separate PR.

BTW I don't think these macro should be replaced with enum. Enum is for set of values. Like in https://github.com/BestImageViewer/geeqie/commit/a4ce95bc11a58700e29c15d324f9ed4d7c27900a#diff-c9276a6068fc1de23fad46678f763188f8a07d8d9327a6513512d47773055dfcR2224
I'd replace these macros and most of enums from a4ce95b with constants of explicit types.

caclark:
@qarkai 
> I'd replace these macros and most of enums from https://github.com/BestImageViewer/geeqie/commit/a4ce95bc11a58700e29c15d324f9ed4d7c27900a with constants of explicit types.

I'm sorry, but I do not understand c++. Perhaps you could give me an opinion about the implementation of this:

Original:
`#define ADVANCED_EXIF_DATA_COLUMN_WIDTH 200
`

Auto-fixed by clang-tidy:
`enum {`
`	ADVANCED_EXIF_DATA_COLUMN_WIDTH = 200`
`};`

I assume that `constexpr` is required, and that constant variable names must be lower-case. So either:
`namespace`
`{`
`constexpr gint advanced_exif_data_column_width = 200;`
`}`
or:
`namespace`
`{`
`constexpr auto advanced_exif_data_column_width = 200;`
`}`
would be OK. Is that so?

qarkai:
> I assume that `constexpr` is required

Not really but it's good for compile-time optimization https://stackoverflow.com/a/13347355/6771534

> So either:
> `namespace`
> `{`
> `constexpr gint advanced_exif_data_column_width = 200;`
> `}`
> or:
> `namespace`
> `{`
> `constexpr auto advanced_exif_data_column_width = 200;`
> `}`
> would be OK. Is that so?

I'd use former, i.e. `constexpr gint` for explicit type.

xsdg:
I'm obviously out of the habit of this codebase, but a lot of C++ codebases I've used have had the habit (just like C) that true constants are still capitalized.

My personal preference (and again, I don't know how much that fits with the current style direction of the codebase) would be for constants (especially global constants) to remain capitalized, because it still makes them easy to distinguish from non-constant variables.

xsdg:
(Also, I agree that it would be better to avoid `auto` for this kind of usecase and use an explicit type)

xsdg:
Every constant from a C-language library will also use `ALL_CAPS` for constants, right?  My personal feeling is that it's important to maintain consistency with that pattern, but again, I'm not the right person to make any conclusive decision.

As for an actual guideline, I think it depends on the intent of the variable.   I think that if it's a constant where the intent is for it to be widely referenced (either within the compilation unit, or more broadly) as a constant, that should always be capitalized.  And generally, any time the original code used `#define`, that's automatically a global constant, and should automatically be capitalized.

That said, if it's a temporary in the middle of a function, and someone's using const or constexpr because it's a good habit, but there is no intent for it to be widely referenced, I think that should generally be up to the author's judgment, defaulting to lower-case.  So I think something like this would be reasonable:
```
namespace {
// This must be capitalized.
constexpr int CYCLE_COUNT = 13;
}  // anonymous namespace

void MyClass::RunCycles() {
  for (int cycle_idx = 0; cycle_idx < CYCLE_COUNT; ++cycle_idx) {
    // For me, either capitalized or lower-case would be fine for both debug_prefix and duration_limit_s,
    // but my preference is lower-case to indicate to code readers that they have limited scope.
    const std::string debug_prefix = absl::StrFormat("[Cycle %d]: ", cycle_idx);
    constexpr int duration_limit_s = 5;

    // I think foo_result and bar_result below _must_ be lower-case.
    const int foo_result = attempt_foo(cycle_idx, duration_limit_s);
    if (foo_result != 0) {
      std::cerr << debug_prefix << "encountered error " << foo_result << " while attempting foo" << std::endl;
    }

    const int bar_result = attempt_bar(cycle_idx, duration_limit_s);
    if (bar_result != 0) {
      std::cerr << debug_prefix << "encountered error " << bar_result << " while attempting bar" << std::endl;
    }
  }
}
```

caclark:
`clang-tidy` `readability-identifier-naming` check has extensive options, all with auto-fix.

xsdg:
From https://clang.llvm.org/extra/clang-tidy/checks/readability/identifier-naming.html

It looks like clang-tidy distinguishes the following types of constants:
- ClassConstant
- Constant
- ConstantMember
- ConstantParameter
- ConstantPointerParameter
- ConstexprVariable
- EnumConstant
- GlobalConstant
- GlobalConstantPointer
- LocalConstant
- LocalConstantPointer
- ScopedEnumConstant
- StaticConstant

I think for what proposal I offered, that would mean `GlobalConstant` and `GlobalConstantPointer` being all caps.  But it's not clear to me that those options provide enough nuance to distinguish between `debug_prefix` (`const`, all caps or lowercase), `duration_limit_s` (`constexpr`, all caps or lowercase), and `foo_result` (`const`, lowercase only)

caclark:
@qarkai  @xsdg  @ anyone

> I'd replace these macros and most of enums from https://github.com/BestImageViewer/geeqie/commit/a4ce95bc11a58700e29c15d324f9ed4d7c27900a with constants of explicit types.

This problem has resurfaced when I ran `clang-tidy --autofix` on some changes:

`src/exif.cc:1294:9: error: macro 'EXIF_DATA_AS_TEXT_MAX_COUNT' defines an integral constant; prefer an enum instead [modernize-macro-to-enum,-warnings-as-errors]`

And the single-value-enum check throws an error.

The opinion of the Internet is:

image

Which way should Geeqie go? If single value enums are to be avoided, I can put NOLINTs wherever appropriate.

qarkai:
> `defines an integral constant; prefer an enum instead`

Not sure why clang-tidy doesn't suggest good old constant/constexpr for single macro.

> The opinion of the Internet is

It also doesn't include `const`/`constexpr` feature.

> Which way should Geeqie go?

I still vote for constants. Specifically `EXIF_DATA_AS_TEXT_MAX_COUNT` could be internal constant of `exif_item_get_data_as_text_full()` function. Something like
```c++
constexpr gint EXIF_DATA_AS_TEXT_MAX_COUNT = 16;
const gint ne = std::min(item->elements, EXIF_DATA_AS_TEXT_MAX_COUNT);
```

caclark:
@qarkai 
If you ask the Internet a question, the answer depends.... AI is not perfect ! :
image

So const/constexpr would be preferable.

PR #1272, opened 2024-03-06
Updated Swedish translation

gusnan:
Updated Swedish po file.


caclark:
Implemented in commit a07a2054b

PR #1277, opened 2024-03-09
Add PanColor to simplify pan functions

qarkai:
Move some defines to cc files.


caclark:
Implemented in commit afe2510

PR #1278, opened 2024-03-09
Simplify util_clip_triangle()

qarkai:


caclark:
Implemented in commit 574939cb9e

PR #1280, opened 2024-03-10
Use p_step to avoid has_alpha comparison in loop

qarkai:


caclark:
Implemented in commit 2a3aaf4f16

PR #1281, opened 2024-03-10
Use std::swap instead of temporary values

qarkai:
Searched for `t`, `tmp`, `temp` variables.


caclark:
Implemented in commit 8c8de657d61

PR #1284, opened 2024-03-11
Fix some GTK deprecation warnings

qarkai:


caclark:
Implemented in commit c7e685513f

PR #1288, opened 2024-03-14
Silence GTK deprecation warnings in champlain-gtk

qarkai:
Also remove unused pragma.


caclark:
Implemented in commit bb9c132df1bb

PR #1290, opened 2024-03-16
Replace deprecated gdk_window_get_pointer

qarkai:
Simplify comparison a bit.


caclark:
Implemented in commit f1f4117c233ef5

PR #1291, opened 2024-03-16
Enable and fix clang-tidy readability-simplify-boolean-expr

qarkai:
Replace some macros with inline functions.


caclark:
Implemented in commit 1c963ff15a

PR #1292, opened 2024-03-16
Convert group_keys to std::array

qarkai:


caclark:
Implemented in commit 4174e0bc

PR #1294, opened 2024-03-17
Fix NetBSD build

0-wiz-0:


src/layout-util.cc
```diff
@@ -24,6 +24,7 @@
 #include <dirent.h>
 #include <sys/wait.h>
 #include <unistd.h>
+#include <sys/wait.h>
```
qarkai, line 27: The same include is on the line 25.

src/misc.cc
```diff
@@ -34,6 +34,7 @@
 #include <memory>
 
 #include <langinfo.h>
+#include <locale.h>
```
qarkai, line 37: Already mentioned on line 29 as `<clocale>`.

qarkai (commented):
@0-wiz-0 Have you tried build with the latest commits?

0-wiz-0:
No, I just built the latest release.
Please make a newer one then :)

PR #1295, opened 2024-03-17
Silence GTK deprecation warning in unused function

qarkai:
Also use G_GNUC_BEGIN/END_IGNORE_DEPRECATIONS macros instead of pragmas.


caclark:
Implemented in commit 90780f290de

PR #1297, opened 2024-03-23
Convert GET_{LEFT,RIGHT}_PIXBUF_OFFSET macro to function

qarkai:
Also cleanup a bit.


caclark:
Implemented in commit 38b1ee1f45d75

PR #1299, opened 2024-03-23
Use g_list_free_full to free draw_queue and draw_queue_2pass

qarkai:


caclark:
Implemented in commit 4a2042605774

PR #1300, opened 2024-03-24
Deduplicate rt_border_draw

qarkai:


caclark:
Implemented in commit b036a1603468f6a9dd

PR #1301, opened 2024-03-24
Cleanup rt_overlay_get_position() and use in rt_draw_cb()

qarkai:


caclark:
Implemented in commit ab812c19295

PR #1304, opened 2024-03-30
Add HtmlBrowser class and replace C array with std::array

qarkai:


caclark:
Implemented in commit 5f971587a658ff

PR #1306, opened 2024-04-01
Simplify ImageSimilarityData comparison

qarkai:
* Run alternate_image_sim_compare_fast() once
* Deduplicate ImageSimilarityData comparison


caclark:
Implemented in commit 2f975870c929

PR #1308, opened 2024-04-02
Replace mround with STL lround

qarkai:


caclark:
Implemented in commit d802d5f5435d6

PR #1310, opened 2024-04-05
Simplify pixels sorting in image_sim_channel_equal()

qarkai:
And move it to anonymous namespace.


caclark:
Implemented in  commit d5648bdf2a006

PR #1311, opened 2024-04-05
Replace custom PI with G_PI

qarkai:


caclark:
Implemented in commit d56bc5ec040a6c

PR #1312, opened 2024-04-05
Remove unused key_press_cb() declaration

qarkai:


caclark:
Implemented in commit 3cb4f5618605c

PR #1319, opened 2024-04-06
Simplify max value search in history-list

qarkai:


caclark:
Implemented in commit 25c115da8ef

PR #1320, opened 2024-04-06
Simplify lists handling in history-list

qarkai:


caclark:
Implemented in commit 52858ec9d409

PR #1321, opened 2024-04-07
Move keyboard_scroll_calc() to layout-util

qarkai:
Convert parameters from pointers to references.
Move Shift processing inside keyboard_scroll_calc().


caclark:
Implemented in commit 425eb031a23ac

PR #1322, opened 2024-04-07
Move some defines from main-defines to modules

qarkai:
Convert to constants.


caclark:
Implemented in commit efcb9615e69

PR #1325, opened 2024-04-08
Drop obsolete GTK_OBJECT_FLOATING

qarkai:


caclark:
Implemented in commit 43f2b53ba8d

PR #1326, opened 2024-04-10
Merge SCROLL_RESET_* macros with PixbufRendererScrollResetType

qarkai:
Rename to ScrollReset.


qarkai:
FreeBSD build fails due to `<mntent.h>` include in `collect-io.cc`.

caclark:
Implemented in commit f91df1b04e096

PR #1327, opened 2024-04-12
Move some types and constants from pan-types to specific modules

qarkai:


caclark:
Implemented in commit b7a3426d7badff8

PR #1328, opened 2024-04-13
Use references in pan_layout_compute() parameters

qarkai:
Simplify related functions.


caclark:
Implemented in commit 2a5c2a898b9914

PR #1329, opened 2024-04-13
Deduplicate PanCacheData destroying

qarkai:


caclark:
Implemented in commit e6ad0f05a

PR #1330, opened 2024-04-13
Simplify pan_item_image_find_size()

qarkai:


caclark:
Implemented in commit a4f88daece63976a8

PR #1331, opened 2024-04-16
Add vf_selection_foreach()

qarkai:
Hide implementation and deduplicate code.
Also simplify layout_menu_write_rotate() a bit.


caclark:
Implemented in commit e57cfab83e6ca

PR #1332, opened 2024-04-17
Deduplicate "Show star rating" menu item creation

qarkai:


caclark:
Implemented in commit 98b2cdbeb4

PR #1333, opened 2024-04-17
Move common click_fd member to ViewFile

qarkai:


qarkai:
@caclark 
[TODO](https://github.com/BestImageViewer/geeqie/blob/master/src/view-file.h#L38) suggests to make union of `ViewFileInfoList` and `ViewFileInfoIcon`.
What do you think of turning them to subclasses of `ViewFile`? It allows to get rid of all `switch (vf->type)`.

caclark:
Implemented in commit fbaa7a3a0fa4d7

qarkai:
@xsdg sorry, I didn’t think to check that this is the name of the active developer.

PR #1335, opened 2024-04-21
Deduplicate ViewFile dnd

qarkai:


caclark:
Implemented in commit 3000a47a87

PR #1336, opened 2024-04-21
Deduplicate ViewFile pop menu functions

qarkai:


caclark:
Implemented in commit 7a46b2160a795

PR #1337, opened 2024-04-21
Prefer internal ViewFile submodule functions

qarkai:
Avoid redundant type checks.


caclark:
Implemented in commit 523a1493c2a7

PR #1338, opened 2024-04-22
Adds API documentation to functions in pixbuf-util.h

xsdg:
Also renames some variables to clarify operation.

Also notes but does not perform some function renames that would dramatically clarify what those functions do.

This change should be a no-op.


caclark:
Implemented in commit 31a812337ef4

PR #1340, opened 2024-04-28
Deduplicate mark_to_selection and selection_to_mark

qarkai:


caclark:
Implemented in commit 844087b69aa3d

PR #1342, opened 2024-04-29
Remove last GTK_STOCK items

qarkai:
Replace `GQ_ICON_COLLECTION` with `PIXBUF_INLINE_COLLECTION`.
Also use `PIXBUF_INLINE_COLLECTION` macro instead of `"gq-icon-collection"` literal.
Replace `GQ_ICON_DND` with `GQ_ICON_DIRECTORY` for `"Dropped list includes folders."` dialog.


caclark:
I am not sure that `folder-pictures` is a good icon for Collections. Collections are not folders - they are files containing references to images. However, I have done some work on the user interface for Collections so that they appear as a pseudo- directory. I think that is more sensible from the user point of view. It will take me a while to finish that.

The icon "folder-pictures" is used by the file manager I use as the link to `$HOME/Pictures`. Therefore it may be confusing to some users.

An icon for Collections appears elsewhere, as shown in the attached screenshot. It would be good to have a consistent style. I have used the icon `./src/icons/gq-collection.png`. The system icon `start-here` is somewhat similar.

A better solution may be to use the standard folder icon with a `g_emblemed_icon_new()`. Unfortunately I could not get that to work. See `./src/view-dir.cc:56`
 
Screenshot_2024-04-30_11-45-04

qarkai:
I'm not insist on `folder-pictures`. Main goal is remove GTK_STOCK to reduce number of warnings. `GQ_ICON_COLLECTION` could be dropped in favor of `PIXBUF_INLINE_COLLECTION`.

caclark:
Implemented in commit cfb3ee3cc7b

Issue #1316, opened 2024-04-06
Panview display corruption on high-dpi with 2x window scaling


xsdg:
**Setup (please complete the following information):**
 - Distribution: [e.g. Ubuntu] Debian
 - Distribution release: [e.g. 20.04] Unstable
 - Geeqie version [geeqie --version]: Geeqie 2.0.1 GTK3 (compiled locally; yes, this is a really old version)

**Describe the bug**
Pan view is unusable in this configuration.  This is likely due to pan-view doing custom rendering that relies on assumptions which no longer hold with window scaling.

**To reproduce**
(I presume) Turn on window scaling and I presume this should break

**Expected behavior**
Pan view should display images correctly, regardless of whether window scaling is enabled or disabled.

**Screenshots**
Default view when entering pan view mode (Timeline mode, default zoom level)
![panview_timeline_default_zoom](https://github.com/BestImageViewer/geeqie/assets/162614/733359bb-0066-4ad0-9120-64ae79e6bcc5)

Switched to grid mode, still default zoom level
![panview_grid_default_zoom](https://github.com/BestImageViewer/geeqie/assets/162614/56ec0b6e-ab10-4201-a40b-585576254607)

Still in grid mode, zoomed out all the way (with Ctrl + scroll-wheel)
![panview_grid_zoomed_out](https://github.com/BestImageViewer/geeqie/assets/162614/eb7735e9-cbc7-4c82-a6dc-ee23377f1fdf)

Grid mode, zoomed in all the way (with Ctrl + scroll wheel)
![panview_grid_zoomed_in](https://github.com/BestImageViewer/geeqie/assets/162614/344e1089-3d59-49b2-a242-41d3931c9f0f)

I use XFCE.  In XFCE settings, showing 2x window scaling setting
![2x_window_scaling_settings](https://github.com/BestImageViewer/geeqie/assets/162614/320c12c6-2465-4b4b-bb20-4cc8c56aa502)

**Additional context**

xsdg:
This looks like the same kinds of artifacts as discussed in https://github.com/BestImageViewer/geeqie/issues/561

Of course, because panview uses a different custom renderer, those changes need to be re-implemented in a different way 😢

Also, I confirmed that this bug continues to manifest with HEAD (commit cac5f75de318c219c16744fb01810ac7fa28e939)

xsdg:
[...]

PR #1343, opened 2024-04-29
Fixes corrupted rendering in PanView on hidpi displays

xsdg:
This is a 95% fix for #1316 .  There is still a bug where rendering corruption will show up if the user zooms all the way out, and will then persist even if they zoom back in again.

The crux of this fix is to account for the hidpi scaling factor in `rt_source_tile_render`.  This PR also adds a significant amount of comments to explain that function's somewhat non-intuitive API, as well as how it operates internally.

A future commit will update the variable names to be more descriptive.

### DIFF
--- a/src/renderer-tiles.cc
+++ b/src/renderer-tiles.cc
@@ -481,10 +481,18 @@ static void rt_hidpi_aware_draw(
 	cairo_t *cr,
 	GdkPixbuf *pixbuf,
 	double x,
-	double y)
+	double y,
+        gboolean should_scale = TRUE)
 {
 	cairo_surface_t *surface;
-	surface = gdk_cairo_surface_create_from_pixbuf(pixbuf, rt->hidpi_scale, nullptr);
+        if (should_scale)
+		{
+		surface = gdk_cairo_surface_create_from_pixbuf(pixbuf, rt->hidpi_scale, nullptr);
+		}
+	else
+		{
+		surface = gdk_cairo_surface_create_from_pixbuf(pixbuf, 1, nullptr);
+		}
 	cairo_set_source_surface(cr, surface, x, y);
 	cairo_fill(cr);
 	cairo_surface_destroy(surface);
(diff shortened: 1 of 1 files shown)


caclark:
Implemented in commits 68cf77cf6bd2 edb32c944f 27acd4cbc 5358cc39531 cbd60bb53ec 9fe066a09e1bf

qarkai:
@xsdg I believe that master branch should not have intermediate work-in-progress commits. Since maintainer doesn't use squash-and-merge strategy would you mind squashing commits for the next pull request?

caclark:
> Echoing https://www.reddit.com/r/learnprogramming/comments/9zgvlk/comment/ea942uh/
"My recommendation is that you only declare multiple per line if they represent
more or less the same thing, and are used together in such a way that they
logically belong to each other."

I am happy if anyone contributes anything, in any way, to the project.

The ability to have a GitHub action check that any pull request or commit complies with a set of standards I see as a great advantage over the old way of no checks - the downside is that there is inflexibility.

In this case here, my option would be to code as, for example:
`gint tile_h;`
`gint tile_w;`
`gint tile_x;`
`gint tile_y;`

This complies with the single line requirement and groups variables together.

My preference is to leave things as they are, and leave the big let-out option as `NOLINT`.

qarkai:
If they "are used together in such a way that they logically belong to each other" then they probably could be combined in struct. For `gint x, y, w, h;` set there is already a `GdkRectangle`. Although it has width/height instead of w/h.

xsdg:
@caclark Just to be clear, the only intent of that comment was to explain that specific commit (From working at a company, I'm in the habit of over-explaining situations where I make a judgment call to bypass automated checks like that)

I agree with you 100% that this automated check makes sense and that it shouldn't be modified.  That said, I think splitting those into separate lines reduces readability because nobody is going to read the declaration for `rx` and then get novel information from a different declaration for `ry` or `rw` or `rh`.  They all always go together.

@qarkai: Yes, a struct is the much better option basically throughout all of the rendering code.  My focus was to fix the bug in time to use panview for a photo job I needed to do over the weekend.  There is _so much_ cleanup that would make sense (both in terms of restructuring and renaming), and I'm planning to put time into it now that the main bug has been resolved.  Thanks for the tip about `GdkRectangle`

PR #1344, opened 2024-05-01
Remove ToolbarButtonData::stock_id

qarkai:
No need to keep it.


caclark:
Implemented in commit a4952a2a6e3d7

PR #1345, opened 2024-05-01
Simplify vflist_get_formatted()

qarkai:
Use GString for result.
Remove redundant with_stars parameter.


caclark:
Implemented in commit cf362d3b5807a

PR #1347, opened 2024-05-08
DO NOT MERGE Creates an early prototype of one way to implement unit …

xsdg:
…tests in Geeqie

Last year, I had tried to untangle the code enough to use googletest in a more standard way (where it only compiles and links against the specific file under test and avoids including the rest of Geeqie), and that didn't work because the Geeqie codebase is so interdependent in its current state.

This approach comes from the opposite direction: just make a version of Geeqie that runs the unit tests instead of actually starting the application.  The next step would be to tie this into some sort of unit test mode flag in meson, and to figure out how to incorporate some C++ unit test framework (I'm most familiar with googletest, so that's my preference).

The meson config should also only ever compile and link the unit test code in the unit test mode, so that code should not be shipped to users as part of releases.

Lastly, it's typical to create parallel `src/` and `tests/` directory hierarchies.  In this case, I think it'll be more straightforward to create a `src/tests/` subdirectory and to put everything there.

```
$./build/src/geeqie
Unit test run function
A condition passed: util_clip_region found overlap
A condition passed: rx as expected
A condition passed: ry as expected
A condition passed: rw as expected
A condition passed: rh as expected
A condition failed: expected to fail
```

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -84,6 +84,8 @@
 #include "ui-fileops.h"
 #include "ui-utildlg.h"
 
+#include "prototype_unit_test.h"
+
 gboolean thumb_format_changed = FALSE;
 static RemoteConnection *remote_connection = nullptr;
 
@@ -1248,7 +1250,23 @@ static void create_application_paths()
 	g_free(path);
 }
 
-gint main(gint argc, gchar *argv[])
+// TODO(xsdg): This define should actually be done conditionally in the build system.
+#define UNIT_TEST_MODE
+
+#ifdef UNIT_TEST_MODE
+	#define MAIN_TEST main
+	#define MAIN_APP _disabled_app_main
+#else
+	#define MAIN_TEST _disabled_test_main
+	#define MAIN_APP main
+#endif
+
+gint MAIN_TEST(gint argc, gchar *argv[])
+{
+	return run_tests();
+}
+
+gint MAIN_APP(gint argc, gchar *argv[])
 {
 	CollectionData *cd = nullptr;
 	CollectionData *first_collection = nullptr;
--- a/src/meson.build
+++ b/src/meson.build
@@ -193,7 +193,10 @@ main_sources = files('advanced-exif.cc',
 'view-dir-tree.h',
 'view-file.h',
 'window.cc',
-'window.h')
+'window.h',  # DO NOT MERGE
+'prototype_unit_test.cc',
+'prototype_unit_test.h',
+)
 
 if conf_data.get('HAVE_DJVU', 0) == 1
     main_sources += files(
--- a/src/prototype_unit_test.cc
+++ b/src/prototype_unit_test.cc
@@ -1,0 +1,71 @@
+/*
+ * Copyright (C) 2024 The Geeqie Team
+ *
+ * Author: Omari Stephens
+ *  
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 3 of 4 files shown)


xsdg:
@caclark @qarkai I'm curious what y'all think of this approach.

xsdg:
Here are some references for how to integrate googletest with meson:

- https://stackoverflow.com/questions/57473395/how-to-set-up-googletest-wtih-meson
- https://groups.google.com/g/mesonbuild/c/q1iBHq0jCQo
- https://github.com/mesonbuild/meson/blob/master/test%20cases/frameworks/2%20gtest/meson.build
- https://bloerg.net/posts/meson-and-google-test/
- https://mesonbuild.com/Unit-tests.html

qarkai:
Is it have to be C++ testing framework? GLib has its own [testing framework](https://docs.gtk.org/glib/testing.html). I'd try it first before adding another dependency.

xsdg:
@qarkai I think that googletest is better designed and has more useful features than the glib test framework.  Plus, I think that testing frameworks are one of the domains where C++'s features (particularly templates) make a _huge_ difference in readability and maintainability.  For example, by not needing to manually register test functions with a separate function call (which is easy to forget to do).  Likewise, as we continue transitioning to code that makes use of C++ features that aren't in C, C-focused frameworks won't be able to handle our needs at all (for example, interacting with code that has constructors and requires `new/delete` rather than `malloc/free`).

So, for example, I rewrite the glib testing sample in googletest.  It's shorter, easier to understand, and can take advantage of C++ features like objects and initializers that aren't available with the glib testing framework.  A lot of the glib testing code is weighed down by boilerplate that googletest doesn't need:
```
#include <clocale>
#include "googletest.h"

constexpr char* TEST_USER_DATA[] = "some-user-data";

class MyObjectFixture : public testing::Test {
  protected:
    MyObjectFixture() : obj(my_object_new()), helper(other_object_new()) {
        my_object_set_prop1(obj, "some_value");
        my_object_do_some_complex_setup(obj, TEST_USER_DATA);
    }

    ~MyObjectFixture() {
        g_clear_object(helper);
        g_clear_object(obj);
    }

    MyObject *obj;
    OtherObject *helper;
};

TEST_F(MyObjectFixture, Test1)
{
    ASSERT_STREQ(my_object_get_property(obj), "initial-value");
}

TEST_F(MyObjectFixture, Test2)
{
    my_object_do_some_work_using_helper(obj, helper);
    ASSERT_STREQ(my_object_get_property(obj), "updated-value");
}

int
main (int argc, char *argv[])
{
    setlocale(LC_ALL, "");

    testing::InitGoogleTest(&argc, argv);
    return RUN_ALL_TESTS();
}
```

And then once we find ways to transition other structs in Geeqie to become real objects, this code becomes even simpler still.  The test logic hasn't changed at all; it's just that more of the boilerplate is done behind the scenes, and so the test logic is easier to read, easier to understand, and easier to modify.
```
#include <clocale>
#include <memory>
#include "googletest.h"

constexpr char* TEST_USER_DATA[] = "some-user-data";

class MyObjectFixture : public testing::Test {
  protected:
    MyObjectFixture() {
        obj.prop1 = "some_value";
        obj.do_some_complex_setup(TEST_USER_DATA);
    }

    std::unique_ptr<MyObject> obj{};
    std::unique_ptr<OtherObject> helper{};
};

TEST_F(MyObjectFixture, Test1)
{
    ASSERT_STREQ(obj.property, "initial-value");
}

TEST_F(MyObjectFixture, Test2)
{
    obj.do_some_work_using_helper(helper);
    ASSERT_STREQ(obj.property, "updated-value");
}

int
main (int argc, char *argv[])
{
    setlocale(LC_ALL, "");

    testing::InitGoogleTest(&argc, argv);
    return RUN_ALL_TESTS();
}
```

xsdg:
Superceded by #1350 .

PR #1348, opened 2024-05-09
Remove unused members in toolbar structs

qarkai:
Widgets are added to containers, no need to keep them in structs.
Remove `UseableToolbarItems`.
Remove redundant casts, merge callbacks.


caclark:
Implemented in commit aeb9f6027818f

PR #1349, opened 2024-05-10
Use more g_list_find_custom()

qarkai:
Instead of manual iteration.


caclark:
Implemented in commit 6203ffc9b45b393

PR #1350, opened 2024-05-10
Creates a Geeqie unit test suite

xsdg:
This incorporates all of the plumbing and infrastructure to be able to build and run unit Geeqie unit tests.

This uses the Googletest framework, as well as Meson pixie dust to be able to use it in-tree on demand.  It operates by modifying the app entrypoint to run unit tests and exit immediately when unit testing is requested.

I took this approach because the Geeqie codebase is so interconnected that it's extraordinarily difficult to actually isolate the code being tested from the rest of the codebase, as might happen in more modern app designs.

Here's a sample run, from a clean tree (`git clean -d -f -x`):
```
$ meson setup -D unit_tests=enabled build
The Meson build system
…
Downloading gtest source from https://github.com/google/googletest/archive/release-1.10.0.zip
Downloading file of unknown size.
Downloading gtest patch from https://wrapdb.mesonbuild.com/v1/projects/gtest/1.10.0/1/get_zip
Download size: 2093
Downloading: ..........

Executing subproject gtest 

gtest| Project name: gtest
gtest| Project version: 1.10.0
gtest| C++ compiler for the host machine: c++ (gcc 12.2.0 "c++ (Debian 12.2.0-14) 12.2.0")
gtest| C++ linker for the host machine: c++ ld.bfd 2.40
gtest| Dependency threads found: YES unknown (cached)
gtest| Dependency threads found: YES unknown (cached)
gtest| Dependency threads found: YES unknown (cached)
gtest| Dependency threads found: YES unknown (cached)
gtest| Build targets in project: 54
gtest| Subproject gtest finished.
…
    unit_tests           : Tests run:
                           YES

  Subprojects
    gtest                : YES

  User defined options
    unit_tests           : enabled

Found ninja-1.11.1 at /usr/bin/ninja

$ meson test -C build -v --suite unit
ninja: no work to do.
ninja: Entering directory `/PATH/TO/geeqie-git/build'
…
[128/128] Linking target src/geeqie
1/1 geeqie:unit / Unit tests RUNNING       
>>> MALLOC_PERTURB_=64 /PATH/TO/geeqie-git/build/src/geeqie --run_unit_tests
――――――――――――――――――――――――――――――――――――― ✀  ―――――――――――――――――――――――――――――――――――――
[==========] Running 4 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 4 tests from ClipRegionTest
[ RUN      ] ClipRegionTest.RegionAContainsRegionB
[       OK ] ClipRegionTest.RegionAContainsRegionB (0 ms)
[ RUN      ] ClipRegionTest.RegionBContainsRegionA
[       OK ] ClipRegionTest.RegionBContainsRegionA (0 ms)
[ RUN      ] ClipRegionTest.PartialOverlapWithBAfterA
[       OK ] ClipRegionTest.PartialOverlapWithBAfterA (0 ms)
[ RUN      ] ClipRegionTest.PartialOverlapWithAAfterB
[       OK ] ClipRegionTest.PartialOverlapWithAAfterB (0 ms)
[----------] 4 tests from ClipRegionTest (0 ms total)

[----------] Global test environment tear-down
[==========] 4 tests from 1 test suite ran. (0 ms total)
[  PASSED  ] 4 tests.
――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
1/1 geeqie:unit / Unit tests OK              0.02s


Ok:                 1   
Expected Fail:      0   
Fail:               0   
Unexpected Pass:    0   
Skipped:            0   
Timeout:            0   

Full log written to /PATH/TO/geeqie-git/build/meson-logs/testlog.txt
```


xsdg:
@qarkai looks like I can't add you as a reviewer also, but I'm curious how this looks to you.

src/tests/pixbuf-util.cc
```diff
@@ -0,0 +1,90 @@
...
+				     500, 500, 1000, 1000,
+				     &rx, &ry, &rw, &rh));
+
+	ASSERT_EQ(500, rx);
+	ASSERT_EQ(500, ry);
+	ASSERT_EQ(500, rw);
+	ASSERT_EQ(500, rh);
+}
+
+TEST_F(ClipRegionTest, PartialOverlapWithAAfterB)
```
qarkai, line 76: Does the same as the `PartialOverlapWithBAfterA`. Need comment if this is intentional.

meson.build
```diff
@@ -611,6 +620,13 @@ ui_sources = []
 subdir('po')
 subdir('plugins')
 
+conditional_unit_test_deps = []
+if conf_data.get('ENABLE_UNIT_TESTS', 0) == 1
+    gtest_proj = subproject('gtest')
```
qarkai, line 625: Is it possible to use system gtest? Like in https://github.com/mesonbuild/meson/blob/master/test%20cases/frameworks/2%20gtest/meson.build

.gitignore
```diff
@@ -57,6 +57,10 @@ Makefile.in
 /src/view_file/.deps
 /src/view_file/.dirstamp
 
+# /subprojects
+/subprojects/packagecache
```
qarkai, line 61: What is it for?

qarkai (changes requested):

.gitignore
xsdg: When meson downloads gtest based on the wrap file, it will create those directories.  Those files should not be checked in:
```
$ls subprojects/*
subprojects/gtest.wrap

subprojects/googletest-release-1.10.0:
appveyor.yml  CMakeLists.txt   googletest/   LICENSE.build   README.md
BUILD.bazel   CONTRIBUTING.md  library.json  meson.build     WORKSPACE
ci/           googlemock/      LICENSE       platformio.ini

subprojects/packagecache:
gtest-1.10.0-1-wrap.zip  gtest-1.10.0.zip
```

src/tests/pixbuf-util.cc
xsdg: This is a mistake; thanks for catching it!

meson.build
xsdg: Added a fallback mechanism to start with system gtest and fallback to the subproject.

src/main.cc
```diff
@@ -84,6 +84,10 @@
 #include "ui-fileops.h"
 #include "ui-utildlg.h"
 
+#if ENABLE_UNIT_TESTS
+#  include "gtest/gtest.h"
```
qarkai, line 88: Does this include style works with system gtest?

src/main.cc
```diff
@@ -663,47 +663,55 @@ static void parse_command_line_for_debug_option(gint argc, gchar *argv[])
...
 			}
 		}
 
-	return ret;
+	return FALSE;
+}
+
+static gboolean search_command_line_for_unit_test_option(gint argc, gchar *argv[])
+{
+	return search_command_line_for_option(argc, argv, "--run_unit_tests");
```
qarkai, line 690: I guess convention for option format is kebab-case. Would you mind to rename it to `--run-unit-tests`?

caclark: https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html
"Option names are typically one to three words long, with hyphens to separate words"

src/main.cc
xsdg: Yes, I tested this.

src/main.cc
xsdg: Fixed; thanks

qarkai:
LGTM!

xsdg:
(I'm not super familiar with the github review workflow.  I saw your lgtm, but it wasn't clear if I needed to click "re-review" for your thing to turn green.  Just slowly figuring this out 😅)

qarkai (approved):

TESTING.md
```diff
@@ -0,0 +1,140 @@
...
+
+After that point, they can be executed with:
+
+```text
+meson test -C build -v --suite unit
+```
+
+Or you can run them by hand by starting geeqie with the `--run-unit-tests`
+argument:
+
```
caclark, line 40: If this is a duplication of the meson option, maybe it would be better to exclude it.
Including the option in bash completions and in the --help output (but only when available) is probably possible but not worth the effort.
Certainly a single-character option is necessary.

TESTING.md
```diff
@@ -0,0 +1,140 @@
+# Geeqie testing and validation patterns
+
```
caclark, line 2: So far as I understand, these tests cannot be run against static functions.
This document:
http://google.github.io/googletest/advanced.html#testing-private-code
describes things that are beyond my comprehension. A recommendation of which path to follow would be good - unless your preference is to not test static functions.

[Maybe you prefer to test only at the public level. Decades ago I did some testing on a project that was itself years old - it worked but required full path coverage for new aircraft. It was an absolute nightmare trying to force a way through each path - even at unit level. I suspect the same would be true for trying to test only at public level on this project]

TESTING.md
```diff
@@ -0,0 +1,140 @@
...
+Runs `clang-tidy` code correctness checks for every source file in the project.
+Note that this will only execute when running from a clone of the Geeqie git
+project.
+
+See `.clang-tidy` and <https://clang.llvm.org/extra/clang-tidy/checks/list.html>
+for more details.
+
+### Single value enum checks
+
+Checks that single-value enums conform to a specified format.
```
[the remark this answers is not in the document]
xsdg: Changed.

TESTING.md
xsdg: So, one thing I had intended that might not be clear is that I wrote `TESTING.md` for developers.  It is not intended for end users to see or pay attention to.

No, it only exists when the meson option is specified, but it is distinct.  It is intended only for triggering unit tests, and I would expect most developers to just let `meson test` handle things.  Here are the four cases:

1. `-Dunit_tests=disabled`, `--run-unit-tests` *not* specified: This is the common case.  Geeqie runs as normal.
2. `-Dunit_tests=disabled`, `--run-unit-tests` *specified*: Geeqie exits immediately with "Unit tests are not enabled in this build."
3. `-Dunit_tests=enabled`, `--run-unit-tests` *not* specified: Geeqie runs as normal, but unit test code is present in the binary.
4. `-Dunit_tests=enabled`, `--run-unit-tests` *specified*: Geeqie runs unit tests and exits immediately.

Because that option will always prevent Geeqie (as an image viewer) from starting, and is not intended for end users, I don't think it makes sense to include in `--help`.  Likewise, I don't think making it easier to specify (like with a single-character option) would be helpful.

TESTING.md
xsdg: I can add some detail about this (plus an example) in a follow-on PR.  Also, I agree that aspiring to 100% test coverage is often not a great goal.  The goal here is to be able to test the things that make sense to test (I'm talking to you, `FileData`)

But in short, what `static` means is `only visible in this file`.  It's the same as anonymous namespaces in C++, `namespace /*no name here*/ {...}`

So one option is to add the unit tests to the implementation file (either directly, or via some weird include, like `#include "tests/inline_blahblah_test.cc"`).  Another option is to add a trampoline object to the file in question, that will make those static methods visible for unit tests.  This could be done with a macro that is only active when `ENABLE_UNIT_TESTS` is defined.

xsdg:
Here's a patch you can `git am` for this pull request.
[00_geeqie_unit_test_suite.patch.txt](https://github.com/BestImageViewer/geeqie/files/15299267/00_geeqie_unit_test_suite.patch.txt)

xsdg:
Merged by hand.

PR #1351, opened 2024-05-11
Creates and uses a helper utility to run tests in a hermetic environment

xsdg:
This avoids interference from/with other processes that might be running on the same host machine (such as the user's own Geeqie processes, or other processes that were started in parallel by Meson), and also ensures that we can clean up the test environment without making any permanent changes (especially important in case of bugs).

## Note: this is stacked on top of #1350, so should be rebased after that PR is merged.


meson.build
```diff
@@ -200,6 +200,15 @@ else
     summary({'execinfo' : ['stacktrace supported:', false]}, section : 'Debugging', bool_yn : true)
 endif
 
+# TODO(xsdg): Also force-enable unit tests in developer mode.
```
caclark, line 203: This comment is an aside to the PR. 
The Ddevel= option is badly named. It should be -Dextended-stacktrace= and should be enabled by default. If the packagers do not want it, that is up to them to disable it. But I think that if a user experiences a crash, the developers will want it. Attached is an example of the output.
I also used that option as a flag for something else - usually a bad idea. If someone clones the repo. for a local build, it is not necessary for the test images to be download. That should be another option.
[stacktrace.txt.gz](https://github.com/BestImageViewer/geeqie/files/15282585/stacktrace.txt.gz)

xsdg: Got it; dropped this TODO from the original commit.

That said, this is clearly super helpful, but I'd wonder how much it might affect binary size, and whether it has an impact on execution speed.

caclark:
I have no more comments on this. It would be good for me if the current Pull Requests were closed, and, whenever, a single Pull Request made.
Just making a `git am` on a `.patch` file makes things easy for me.

xsdg:
Here's a patch you can `git am` for this pull request.  It should be applied after #1350.
[01_geeqie_test_isolation.patch.txt](https://github.com/BestImageViewer/geeqie/files/15299277/01_geeqie_test_isolation.patch.txt)

xsdg:
Merged by hand.

PR #1352, opened 2024-05-11
Add MetadataCacheEntry for fd->cached_metadata

qarkai:


caclark:
Implemented in commit 65e0295d7bd7

PR #1353, opened 2024-05-11
Remove ToolbarButtonData

qarkai:
Use simple gchar*.
Also simplify toolbar_apply.


caclark:
Implemented in commit 5301bf1e5ad8f

PR #1354, opened 2024-05-14
Simplify util_clip_region()

qarkai:
* use gdk_rectangle_intersect()
* convert pointers to references

@xsdg 
Is it worth to replace parameters with `GdkRectangle`s, making `util_clip_region` wrapper of `gdk_rectangle_intersect`?

### DIFF
--- a/src/pixbuf-util.cc
+++ b/src/pixbuf-util.cc
@@ -395,30 +395,23 @@ GdkPixbuf *pixbuf_fallback(FileData *fd, gint requested_width, gint requested_he
  */
 
 gboolean util_clip_region(gint x, gint y, gint w, gint h,
-			  gint clip_x, gint clip_y, gint clip_w, gint clip_h,
-			  gint *rx, gint *ry, gint *rw, gint *rh)
+                          gint clip_x, gint clip_y, gint clip_w, gint clip_h,
+                          gint &rx, gint &ry, gint &rw, gint &rh)
 {
-	// Ensures that clip region and main region have some overlap (they aren't
-	// completely disjoint).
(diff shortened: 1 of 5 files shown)


xsdg:
I think this is a good enough chunk of work to keep in its own PR.

I think it would be good to switch to using GdkRectangle throughout the rendering system, but that should be its own PR.  At that point, I would just drop `util_clip_region` and have the callsites call `gdk_rectangle_intersect` directly.

src/pixbuf-util.cc
```diff
@@ -395,30 +395,23 @@ GdkPixbuf *pixbuf_fallback(FileData *fd, gint requested_width, gint requested_he
...
-	// completely disjoint).
-	if (clip_x + clip_w <= x ||  /* assert(x < clip_right) && */
-	    clip_x >= x + w ||       /* assert(clip_x < right) && */
-	    clip_y + clip_h <= y ||  /* assert(y < clip_bottom && */
-	    clip_y >= y + h)         /* assert(bottom < clip_y) */
+	GdkRectangle main{x, y, w, h};
+	GdkRectangle clip{clip_x, clip_y, clip_w, clip_h};
+	GdkRectangle r;
+
+	gboolean res = gdk_rectangle_intersect(&main, &clip, &r);
```
xsdg, line 405: Instead of `res`, I think it'd be more helpful to name the variable for what it actually represents.  So maybe `found_intersection`?  Also, as a habit, this can be `const`

qarkai: Renamed to `rectangles_intersect`.

xsdg: lgtm

caclark:
Implemented in commit 6890489990

PR #1355, opened 2024-05-15
Use gdk_rectangle_intersect() in pan-view

qarkai:


qarkai:
@xsdg could you review? Don't know how to add reviewer via GitHub interface.

src/pan-view/pan-view.cc
```diff
@@ -368,42 +368,35 @@ static gboolean pan_window_request_tile_cb(PixbufRenderer *pr, gint x, gint y,
 	auto pw = static_cast<PanWindow *>(data);
 	GList *list;
 	GList *work;
-	gint i;
+	const GdkRectangle rect{x, y, width, height};
```
xsdg, line 371: I might modify these variable names to make this code a little clearer.  Here's one option:
`rect` -> `request_rect`
`clip` -> `pan_grid_rect`

Also, I'm _pretty_ sure, that this code is extra confusing because it's actually drawing lines and not rectangles (hence why the width and height are 1px).  Probably not something to poke at in this PR, but just an fyi that there's a lot being done here that's really counter-intuitive.  If you look at pan view, I believe these are the lines that create the grid pattern in the background.

src/pan-view/pan-view.cc
```diff
@@ -807,43 +799,30 @@ static void pan_grid_build(PanWindow *pw, gint width, gint height, gint grid_siz
 			}
 		}
 
-	work = pw->list;
-	while (work)
+	for (GList *work = pw->list; work; work = work->next)
```
xsdg, line 802: This pattern is a nice simplification 👍🏽

xsdg (approved):

src/pan-view/pan-view.cc
qarkai: Applied renaming suggestions.

`pixbuf_draw_line` also uses clipping rectangle. It's not obvious to me how to use it instead of `pixbuf_draw_rect_fill`.

caclark:
Implemented in commit db1325591f9

PR #1356, opened 2024-05-16
Use util_clip_triangle() in pan_item_tri_new()

qarkai:
Remove unused parameters in pan_item_tri_new().


caclark:
Implemented in commit f12d68b7a20

PR #1357, opened 2024-05-17
Use GdkPoint in PAN_ITEM_TRIANGLE coord

qarkai:


src/pan-view/pan-item.cc
```diff
@@ -320,37 +316,37 @@ gint pan_item_tri_draw(PanWindow *, PanItem *pi, GdkPixbuf *pixbuf, PixbufRender
...
-				     coord[2] - x, coord[3] - y,
-				     coord[4] - x, coord[5] - y,
-				     pi->color.r, pi->color.g, pi->color.b, pi->color.a);
+		                     rx - x, ry - y, rw, rh,
+		                     coord[0].x - x, coord[0].y - y,
+		                     coord[1].x - x, coord[1].y - y,
+		                     coord[2].x - x, coord[2].y - y,
+		                     pi->color.r, pi->color.g, pi->color.b, pi->color.a);
 
 		if (pi->border & PAN_BORDER_1)
```
xsdg, line 327: This a borderline judgment call (so take this more as an observation than an actual suggestion), but personally, I feel like using a lambda here to reduce code duplication would be a win.  It's more of a win with single-line if statements, but I think it's still easier to understand even with the extra braces.  Something like:
```
const auto draw_line = [&](const GdkPoint &start, const GdkPoint &end)
    {
    pixbuf_draw_line(..., start.x - x, start.y - y, end.x - x, end.y - y, ...);
    };

if (pi->border & PAN_BORDER_1) draw_line(coord[0], coord[1]);
if (pi->border & PAN_BORDER_2) draw_line(coord[1], coord[2]);
if (pi->border & PAN_BORDER_3) draw_line(coord[2], coord[0]);

```

qarkai: Added lambda.

PR #1358, opened 2024-05-17
Use GdkRectangle in PanGrid

qarkai:
@xsdg any naming suggestions?

### DIFF
--- a/src/pan-view/pan-view.cc
+++ b/src/pan-view/pan-view.cc
@@ -74,10 +74,7 @@ namespace
 {
 
 struct PanGrid {
-	gint x;
-	gint y;
-	gint w;
-	gint h;
+	GdkRectangle rect;
 	GList *list;
 };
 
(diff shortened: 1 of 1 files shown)


src/pan-view/pan-view.cc
```diff
@@ -967,28 +961,23 @@ static GList *pan_layout_intersect_l(GList *list, GList *item_list,
...
+		const auto *rect = static_cast<const GdkRectangle *>(user_data);
 
-		if (x < pg->x || x + width > pg->x + pg->w ||
-		    y < pg->y || y + height > pg->y + pg->h)
-			{
-			pg = nullptr;
-			}
-		}
+		GdkRectangle intersection;
+		return (gdk_rectangle_intersect(&pg->rect, rect, &intersection) && gdk_rectangle_equal(rect, &intersection)) ? 0 : 1;
```
xsdg, line 972: I would split this across multiple lines just for legibility.  (Also, just to note this for future style conversations, this is a place where I would want to use a single-line `if (...) return 1;` instead of adding the braces)
```
GdkRectangle intersection;
if (!gdk_rectangle_intersect(...))
    {
    return 1;
    }
return gdk_rectangle_equal(...)
```

xsdg (approved):
Naming seems fine.  Only note is ternary line I commented on.

xsdg (approved):

caclark:
Implemented in commit bb01893bd8de5e

PR #1361, opened 2024-05-20
Use gdk_rectangle_intersect() in pan-item

qarkai:
Also removed unused variable in `pan_flower_build()`.

@xsdg could you review please?


caclark:
Implemented in commit 55c9be8091

PR #1362, opened 2024-05-20
Replace pr_clip_region() with util_clip_region()

qarkai:


caclark:
Implemented in commit 465d70f79c0

PR #1363, opened 2024-05-21
Use gdk_rectangle_intersect() in pixbuf-renderer

qarkai:


caclark:
Implemented in commit 43d77a58571

PR #1364, opened 2024-05-22
Use gdk_rectangle_intersect() in renderer-tiles

qarkai:


caclark:
Implemented in commit 00fbaf0ef845

PR #1367, opened 2024-05-25
Use GdkPoint in triangle functions

qarkai:
Simplify `util_clip_triangle()` and `pixbuf_draw_triangle()`.


caclark:
Implemented in commit 26bed6d3e

PR #1368, opened 2024-05-26
Use GdkRectangle for clipping region in pixbuf_draw_*() functions

qarkai:
Add `pixbuf_clip_region()` to deduplicate code.


caclark:
Implemented in commit 03cdfb55e8a6e5

PR #1370, opened 2024-05-28
Fix pixbuf_draw_fade_radius()

qarkai:
Distinguish the red param from the radius temporary variable.


caclark:
Implemented in commit 33030d86f538

PR #1372, opened 2024-05-29
Fix pixbuf_draw_fade_linear()

qarkai:
Ignore distance from s greater than border.


caclark:
Implemented in commit 56d997a

PR #1373, opened 2024-05-30
Deduplicate pixbuf_draw_rect_fill()

qarkai:


caclark:
Implemented in commit 71a75d233820

PR #1374, opened 2024-05-30
Partial update for russian translation

dunmaksim:
Hello!

This is a partial update for russian translation.


caclark:
Implemented in commit 29b4a5a91214c2f90

Issue #1369, opened 2024-05-28
Refactor FileData to be more modular and more testable


xsdg:
Currently, FileData is implemented as `filedata.h` (295 lines) and `filedata.cc` (3,581 lines).  `filedata.h` declares 102 functions, and `filedata.cc` also includes 57 static functions.

The purpose of this bug is to lay out the desired end state, since it will take jumping through some hoops to actually get there from how things are implemented right now, while keeping everything functioning during the refactor.

The ideal end state has the following traits:
1) Fundamentally distinct groups of functionality are split into separate implementation files.  An old prototype of this refactor created `changeinfo.cc`, `core.cc`, `filelist.cc`, `filter.cc`, `metadata.cc`, `sidecar.cc`, `sidecar_change_info.cc`, and `util.cc`.  That seems like a reasonable place to start.
2) FileData becomes a class, in order to take advantage of C++ class access control features.
3) static functions are generally replaced with protected functions.  There may be some functions that still make sense to live in an anonymous namespace, but in general, functions should be unit-testable.
4) A shim layer is created to translate between the historical C-style FileData API (things like `file_data_new_dir(const gchar *path_utf8)`) and a more modern C++-style API (something like `FileData::new_for_dir(const gchar *path_utf8)`
5) Globals (like `static GHashTable *file_data_pool`) should at the very least become protected members (so that they're no longer accessible outside of the `FileData` class), and should ideally be encapsulated in some way, so that unit tests don't interfere with each other and to make it easier to make things thread-safe.

xsdg:
For general reference, the prototype implementation was created in this sequence of commits, from a few years ago.  I did get geeqie to compile and run with these changes.
1) [Roughly splits `filedata.c` into multiple files that are grouped around general functionalities](https://github.com/BestImageViewer/geeqie/commit/34640b262a18efc0efcf24eceddf5da887004ae6)
2) [Starts trying to get this to compile and link.](https://github.com/BestImageViewer/geeqie/commit/b69636560b58adca893b655345e73d1fee64c8dd)
3) [Comments out "static" specifier for functions.](https://github.com/BestImageViewer/geeqie/commit/1dde26c75c0ffca5e2d93436c430b95f4b4cb120)
4) [Adds the "FileData::" class specifier to all functions, to turn them into class methods](https://github.com/BestImageViewer/geeqie/commit/45a9e0f9a70443e33575e5eeaf861c9ba43b6562)
[...]

PR #1375, opened 2024-05-31
Turns FileData into a C++-style class (phase 1 of #1369)

xsdg:
See #1369 for more context, but at a high level, this is a step towards being able to use C++-style fine-grained access control on the class methods, functions, and members to clearly define a public FileData API, as well as internal APIs between the different FileData subcomponents.  Furthermore, this will allow us to test the internal implementations (especially around refcounting) without making those internal implementations part of the public API.


xsdg:
@qarkai @caclark I tried to include the minimum number of commits that still keeps mechanical changes separate from changes that actually need more thought to review.  Also, sadly, the 2-step copy procedure (from https://devblogs.microsoft.com/oldnewthing/20190919-00/?p=102904 ) is the only way I could find to retain file history across a copy, and it takes up 3 commits 😢

This is a big change, but I've tried to make sure everything is being run through tests and static analysis, and that the sequence of changes is fairly logical and easy to understand/review.

caclark:
@xsdg - or anyone else...

Sorry, I have more problems with git. This pull request conflicts with the latest Russian translation.
What is the correct way to resolve the conflict?

Thanks

src/filedata.cc
```diff
@@ -21,3561 +21,565 @@
...
-		*d = *s;
-		s++;
-		d++;
-		i--;
-		}
-	*d = '\0';
-
-	g_free(a);
-	return b;
+	return FileData::text_from_size(size);
```
qarkai, line 26: Not sure `text_from_*` belongs to `FileData` but it's better to leave it there for now.

src/filedata.h
```diff
@@ -132,8 +132,182 @@ struct FileData {
...
+
+#ifdef DEBUG_FILEDATA
+	FileData *file_data_ref(const gchar *file = __builtin_FILE(), gint line = __builtin_LINE());
+	void file_data_unref(const gchar *file = __builtin_FILE(), gint line = __builtin_LINE());
+#else
+	FileData *file_data_ref();
+	void file_data_unref();
+#endif
+
+	void file_data_lock(FileData *fd);
```
qarkai, line 168: Is removing `FileData` parameter from non-static methods planned in the next steps?

qarkai:
> @xsdg - or anyone else...
> 
> Sorry, I have more problems with git. This pull request conflicts with the latest Russian translation. What is the correct way to resolve the conflict?
> 
> Thanks

Usually author of PR should resolve conflicts and update PR.

caclark:
If I run `patch -p1 < 1375.diff` and then run `./scripts/test-all.sh` I get the error:

`../src/search.cc:3828:19: error: redundant explicit casting to the same type 'FileData *' as the sub-expression, remove this casting [readability-redundant-casting,-warnings-as-errors]`
` 3828 |                 file_data_unref((FileData *)mfd->fd);`
`      |                                 ^~~~~~~~~~~~`
`../src/search.cc:272:12: note: source type originates from referencing this member`
`  272 |         FileData *fd;`
`      |         ~~~~~~~~~~^`

I do not understand why the error did not show when the pull request tests ran on GitHub.

src/filedata.h
xsdg: Yeah.  Going to drop that, and also drop the C-style "file_data_" name prefix.  So the end state for this method is `void lock();`

xsdg:
> If I run `patch -p1 < 1375.diff` and then run `./scripts/test-all.sh` I get the error:
> 
> `../src/search.cc:3828:19: error: redundant explicit casting to the same type 'FileData *' as the sub-expression, remove this casting [readability-redundant-casting,-warnings-as-errors]` ` 3828 | file_data_unref((FileData *)mfd->fd);` ` | ^~~~~~~~~~~~` `../src/search.cc:272:12: note: source type originates from referencing this member` ` 272 | FileData *fd;` ` | ~~~~~~~~~~^`
> 
> I do not understand why the error did not show when the pull request tests ran on GitHub.

You're probably using a newer version of clang.  This check was added very recently: \
https://github.com/llvm/llvm-project/issues/67534

(Incidentally, this is one of the big downsides of things that behave like `-Wall` -- new compiler versions can cause code to stop compiling that used to compile without issue.)  My preference is to fix this in a separate PR.

src/filedata.cc
xsdg: Agreed.  I think there are a lot of util functions that have ended up scattered around the codebase, and refactors like this one will make it easier to spot them and to move them to more appropriate homes.  I think it'll be best to evaluate that once filedata.cc is split into separate files, though.

xsdg:
Rebased to pick up ru.po update.  Should patch cleanly now.

caclark:
Implemented in commits 43c3a767a97b 7a5da6e9232f22cf 0c26f3819bd97 e5648fc1579b555 9414f22efcafd3b 0832137ce92c5 4978b5bc6d3121d 1ebf122bce5e6 73765be18987d 13e8d9f487d9594 edf7b46fcd9ec ab671bdfb1bed 18183df6196d 535c3054a30f 0f50e13335

src/filedata.h
```diff
@@ -74,7 +70,11 @@ struct FileDataChangeInfo {
 	gboolean regroup_when_finished;
 };
 
-struct FileData {
+class FileData {
```
qarkai, line 73: Reminder for the next PR. Convert forward declarations from `struct FileData` to `class FileData`.

PR #1377, opened 2024-05-31
Simplify pixels filling a bit

qarkai:


caclark:
Implemented in commit 37a5136aad

PR #1378, opened 2024-06-01
Remove unused GtkWidget parameter from pixbuf_draw_layout()

qarkai:


caclark:
Implemented in commit 43c66d6a9c9

PR #1382, opened 2024-06-03
Creates an initial FileData unit test

xsdg:
The utility of this is currently limited by the inability to inject dependencies or otherwise mock function calls within the FileData implementation.  This will improve as the FileData refactor proceeds, both by replacing static methods with more testable alternatives, as well as by creating more formal internal APIs that can be mocked more easily.


caclark:
Implemented in commit 756bff0f78edf1

PR #1383, opened 2024-06-04
Drop util_clip_region()

qarkai:
Use gdk_rectangle_intersect() directly.
Also deduplicate pixbuf_draw_shadow().


caclark:
Implemented in commit a22c0efc23195

PR #1384, opened 2024-06-05
Moves FileList functions to their own file/class (Phases 3/4 from #1369)

xsdg:
This also creates a protected FileData API, which includes methods that originally existed as static functions within `filedata/filedata.cc`.  Those protected methods are now accessible to FileData and its child classes, without being made accessible to code outside of the FileData class.

This is part of the #1369 refactor (phases 3 and 4)

### DIFF
--- a/src/filedata.h
+++ b/src/filedata.h
@@ -305,35 +305,41 @@ class FileData::FileList
 	friend class FileData;  // Allows FileData to access protected API.
 
 	// Globals.
-	static SortType filelist_sort_method;
-	static gboolean filelist_sort_ascend;
-	static gboolean filelist_sort_case;
+	static SortType sort_method;
+	static gboolean sort_ascend;
+	static gboolean sort_case;
 
     public:
-	static gint filelist_sort_compare_filedata(FileData *fa, FileData *fb);
-	static gint filelist_sort_compare_filedata_full(FileData *fa, FileData *fb, SortType method, gboolean ascend);
-	static GList *filelist_sort(GList *list, SortType method, gboolean ascend, gboolean case_sensitive);
-	static GList *filelist_sort_full(GList *list, SortType method, gboolean ascend, gboolean case_sensitive, GCompareFunc cb);
-	static GList *filelist_insert_sort_full(GList *list, gpointer data, SortType method, gboolean ascend, gboolean case_sensitive, GCompareFunc cb);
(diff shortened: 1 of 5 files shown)


xsdg:
@qarkai definitely going to want your thoughts on this

src/filedata/filelist.cc
```diff
@@ -0,0 +1,537 @@
...
+gboolean FileData::FileList::sort_ascend = TRUE;
+gboolean FileData::FileList::sort_case = TRUE;
+
+/*
+ *-----------------------------------------------------------------------------
+ * handling sidecars in filelist
+ *-----------------------------------------------------------------------------
+ */
+
+static GList *filter_out_sidecars(GList *flist)
```
qarkai, line 73: Would you mind moving static functions to anonymous namespace?

src/filedata.h
```diff
@@ -293,16 +278,66 @@ class FileData {
...
+	friend class FileData;  // Allows FileData to access protected API.
+
+	// Globals.
+	static SortType sort_method;
+	static gboolean sort_ascend;
+	static gboolean sort_case;
+
+    public:
+	static gint sort_compare_filedata(FileData *fa, FileData *fb);
+	static gint sort_compare_filedata_full(FileData *fa, FileData *fb, SortType method, gboolean ascend);
```
qarkai, line 314: Not sure these methods belong to `FileList`. But since it's mechanical moving/renaming I'm ok for now. TODO note would be great.

xsdg: Yeah, I agree that it doesn't belong here.  I don't know exactly what order I'll be creating the subclasses in, but there will be a `util.cc` file which will collect a lot of these.

src/filedata/filelist.cc
xsdg: This is a good catch; thanks.  I saw that there are only 4 static functions left, so I just added them to the protected API.

src/filedata.h
xsdg: Oh, I see what happened.  The implementation of those functions depends on the `filelist_sort_method` globals, which I had moved to this class.  I have an idea for how to manage those globals better, but it'll take more work.

qarkai (approved):
LGTM

caclark:
Implemented in commits d2eb115afd5 a72d5ce8b

PR #1385, opened 2024-06-07
Use GdkRectangle in pixbuf_draw_rect_fill()

qarkai:


caclark:
Implemented in commit e40b9e5efd8b8e6

PR #1386, opened 2024-06-08
Return GdkRectangle from util_triangle_bounding_box()

qarkai:
Rename from util_clip_triangle().


caclark:
Implemented in commit e41669bbf

PR #1389, opened 2024-06-09
Convert PanTextAlignment to class

qarkai:
Convert `pan_text_alignment_*` functions to methods.
Add private internal `struct Items`.
Replace `GLists` with `vector` of `Items`.


caclark:
Implemented in commit 3b09f01bc

PR #1390, opened 2024-06-09
Return GdkRectangle from rt_overlay_get_position()

qarkai:


caclark:
Implemented in commit 86138e3e78166e

PR #1391, opened 2024-06-09
Use GdkRectangle in rt_overlay_draw()

qarkai:


caclark:
Implemented in commit f6c084d236d8

PR #1392, opened 2024-06-09
Use GdkRectangle in rt_border_draw()

qarkai:


caclark:
Implemented in commit c3f40e423f0f

PR #1393, opened 2024-06-11
Use GdkRectangle in RendererFuncs::invalidate_region()

qarkai:


caclark:
Implemented in commit 8597fe1c1b9

PR #1395, opened 2024-06-12
Use GdkRectangle in rt_tile_get_region()

qarkai:
Simplify a bit.


caclark:
Implemented in commit fa12470d08ef

PR #1397, opened 2024-06-15
Convert pointers to references in image getters

qarkai:
Simplify related functions.

### DIFF
--- a/src/image.cc
+++ b/src/image.cc
@@ -1977,9 +1977,7 @@ void image_background_set_color_from_options(ImageWindow *imd, gboolean fullscre
 	image_background_set_color(imd, color);
 }
 
-void image_color_profile_set(ImageWindow *imd,
-			     gint input_type,
-			     gboolean use_image)
+void image_color_profile_set(ImageWindow *imd, gint input_type, gboolean use_image)
 {
 	if (!imd) return;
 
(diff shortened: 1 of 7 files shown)


caclark:
Implemented in commit b016089c8d579eb3

caclark:
@qarkai 
I have been looking at creating a snap for Geeqie. That involves using `24.04` as a base.
I get an error on  line 751 in `command-line-handling.cc` which originated in line 828 of `remote.cc` in this commit. 

ChatGPT gives this explanation:

-------
../src/src/command-line-handling.cc:751:65: error: ‘abs’ is not a member of ‘std’
...
note: ‘abs’ declared here: /usr/include/stdlib.h

That means the file uses std::abs(...) but doesn’t include the C++ header that puts the overloads into std:::

    std::abs(int) lives in <cstdlib>

    std::abs(double/float/long double) live in <cmath>

On some builds it “works by accident” via transitive includes; in the snap GNOME SDK it doesn’t.
Quick fixes
Option A (preferred): patch the source to include `<cstdlib>`

Add at the top of src/command-line-handling.cc:
#include `<cstdlib>`   // for std::abs

If those abs calls are on floating values anywhere, also add:

#include `<cmath>`     // for std::abs(double/float)

#include `<cstdlib>`   // for std::abs

----------
 Is this understandable to you?

PR #1398, opened 2024-06-15
Simplify name_sidecars constructing in vficon_cell_data_cb()

qarkai:


caclark:
Implemented in commit 04d112536fe8750

PR #1399, opened 2024-06-15
Use G_STMT_START/G_STMT_END

qarkai:
Simplify print_term().

### DIFF
--- a/src/debug.h
+++ b/src/debug.h
@@ -49,24 +49,26 @@ gint required_debug_level(gint level);
 const gchar *get_exec_time();
 void init_exec_time();
 
-#define DEBUG_N(n, ...) do \
-				{                                                      \
-				_Pragma("GCC diagnostic push")                         \
-				_Pragma("GCC diagnostic ignored  \"-Wformat\"")        \
-				gint debug_level = get_debug_level(); \
-				if (debug_level >= (n)) 	\
-					{ 		\
-					if (debug_level != 1) \
-						{ \
-						log_domain_print_debug(DOMAIN_DEBUG, __FILE__, __LINE__, __func__, __VA_ARGS__);\
-						} \
-					else \
-						{ \
-						log_domain_printf(DOMAIN_DEBUG, __VA_ARGS__); \
-						} \
-					}                                                  \
-				_Pragma("GCC diagnostic pop")                          \
-				} while (0)
+#define DEBUG_N(n, ...) \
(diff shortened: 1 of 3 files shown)


caclark:
Implemented in commit 2374e04295d

xsdg:
In future: it would be helpful for the commit (or PR) description to include what `G_STMT_START/END` actually do, so that people who aren't familiar wouldn't necessarily need to do a separate search to understand why this change makes sense.

Even if you didn't want to write out an explanation in your own words, linking to an explanation like https://stackoverflow.com/questions/23209951/what-is-are-g-stmt-start-and-g-stmt-end would be very helpful.

qarkai:
Sorry I didn't realize that GLib macros could be unfamiliar.
Basically `G_STMT_*` can use compiler's extensions and quirks.

caclark:
I make this comment solely as an observation about how these days one has become accustomed to be able to find information about anything - but sometimes that does not happen.

G_STMT_START is described in the Glib document:
`file:///usr/share/doc/libglib2.0-dev/glib/index.html`

But type G_STMT_START in the search box, and there is no hit.
However if you know what you want, you can go to:
`file:///usr/share/doc/libglib2.0-dev/glib/macros.html?do-not-autocomplete=ABS#title`

Similarly with Devhelp (which is what I use) - the search box does not provide a hit, but if you scroll down through the Glib documents, the information is there (again, only if you know where to look).

If a reference is required, I think that a link to the official source data is preferable.

But in the end any documentation is better than none - in spite of me not doing much in that area.

qarkai:
https://docs.gtk.org/glib/macros.html#miscellaneous-macros
I used `site:docs.gtk.org G_STMT_START` in Google.

xsdg:
> Sorry I didn't realize that GLib macros could be unfamiliar. Basically `G_STMT_*` can use compiler's extensions and quirks.

@qarkai I feel like one of your strengths is understanding all of the features available in a library, figuring out where they fit best, and then using them to make the codebase better.  That's been _super_ helpful throughout all of the time you've been working on Geeqie, and I wanted to verbalize that.

I would just emphasize that this is a particular strength of yours, and sometimes other folks will have trouble following changes that might seem obvious.  That's one reason why adding descriptions to "obvious" changes can be really important -- it helps people understand the change who might have strengths in other areas, or who just might be less familiar with the features of that library.

PR #1401, opened 2024-06-17
Simplify collection_info_list_to_dnd_data()

qarkai:
Use `GString` to collect result instead of `GList` of `gchar *`.
Convert pointer to reference.
Add const.


caclark:
Implemented in commit 7b249f36de57f7147c63d34689c954ddf6b93050

PR #1403, opened 2024-06-19
Simplify collection_from_dnd_data()

qarkai:
Use g_strsplit() instead of manual parsing.


caclark:
Implemented in commit 41b0af98a7b3f41b8b73d16b2325f8c7f0140c13

PR #1409, opened 2024-06-25
Move PanCacheData and related functions to pan-view

qarkai:
Rename `pan_item_image_find_size()` to `pan_cache_get_image_size()` and simplify a bit.


caclark:
Implemented in commit 64ec1aa07511aadafcca6b16eb51625f6943bc24

PR #1412, opened 2024-06-27
Fixes a thumbnail permissions bug introduced in a4ce95bc1

### LINKED ISSUE #1410: No read permissions on thumbnails [patch included]
 - Distribution: gentoo
 - Distribution release: master
 - Geeqie version [geeqie --version]: 2.4

**Describe the bug**
In 2.4 geeqie creates thumbnails with permission 000 for some reason. As a result, nothing, not even geeqie can read them. Otherwise thumbnails work the same as previously.

**To reproduce**
Build **without** exif
Set thumbs to ~/.cache/thumbnails/

xsdg:
That bug caused thumbnails to be created with no permissions (0000), instead of U+RW permissions (0600).

Unfortunately, I don't think there's a good way to auto-fix this issue without potentially adjusting permissions that Geeqie didn't originally set, which wouldn't be appropriate.

Fixes #1410

### DIFF
--- a/src/thumb-standard.cc
+++ b/src/thumb-standard.cc
@@ -369,7 +369,7 @@ static void thumb_loader_std_save(ThumbLoaderStd *tl, GdkPixbuf *pixbuf)
 					  NULL);
 		if (success)
 			{
-			chmod(pathl, (tl->cache_local) ? tl->source_mode : S_IRUSR & S_IWUSR);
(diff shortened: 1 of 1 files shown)


caclark:
Implemented in commit b4f4d1f03da37ec8068011e6351b1b1803cc654e

PR #1413, opened 2024-06-28
Reduce number of reinterpret_cast

qarkai:
Use GObject macros.
Fix types in callbacks.
Remove redundant calls.

### DIFF
--- a/src/bar.cc
+++ b/src/bar.cc
@@ -608,7 +608,7 @@ void bar_clear(GtkWidget *bar)
 
 	list = gtk_container_get_children(GTK_CONTAINER(bd->vbox));
 
-	g_list_free_full(list, reinterpret_cast<GDestroyNotify>(remove_child_from_parent));
+	g_list_free_full(list, remove_child_from_parent);
 }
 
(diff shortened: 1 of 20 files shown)


caclark:
Part of this pull request moves function description and parameter documentation from the .cc file to the .h file.
There is some discussion on the Internet about which method is correct.
Doxygen picks up the description in either location and displays it with the code, as the attached screenshot of  documentation in the .h file shows.

I now see that there is some documentation in .h files made long ago by John Ellis in the first commits. I did not realise it was there, because I do not read the .h files.

When I am looking at the code I can, via a few keystrokes, get to the documentation either in the .h files or in Doxygen, but of course there is no guarantee that there is any documentation (mostly there is not). Having to switch between the documentation and the code seems unnecessarily tedious.

My preference is for the documentation to be in the .cc files. That seems to be the most useful location for it.

Screenshot_2024-06-29_12-01-54

qarkai:
Reverted docs.

caclark:
Implemented in commit 80990eb37b4fced19701f

PR #1416, opened 2024-06-29
Deduplicate FileData Func types

qarkai:


caclark:
Implemented in commit e12914a6b803e01fb0ca25afe904c4344972640a

PR #1417, opened 2024-06-30
Simplify bar_pane_gps_write_config()

qarkai:
Use `write_int_option()` to avoid redundant memory allocations.
Deduplicate latitude/longitude options writing.
Remove intermediate `h` variable, use `pgd->height` directly.


caclark:
Implemented in commit 8a5300eb9980d0db5211f347c00b9fd7c62165b0

PR #1420, opened 2024-07-01
Creates a FileDataRef RAII object for holding a FileData reference.

### LINKED ISSUE #1415: Create an RAII `FileDataRef` object.
[See here](https://stackoverflow.com/questions/2321511/what-is-meant-by-resource-acquisition-is-initialization-raii) for a quick explanation of RAII.

In short, this would make it easy to `::file_data_ref` an `fd` for the scope of a particular function, and to guarantee that `::file_data_unref` is called on the `fd` when the function is exited, without requiring a separate `::file_data_unref` call to be added to every potential exit point.

This works by creating a locally-scoped object that calls `::file_data_ref` in its constructor (so when the object is created, `::file_data_ref` is called), and then `::file_data_unref` in its destructor (so when the object is destroyed, `::file_data_unref` is called).  Because the object is declared on the stack, it will be created by the compiler IFF that line of code is run, and it will be destroyed IFF it goes out of scope.

xsdg:
Fixes #1415

### DIFF
--- a/src/filedata.h
+++ b/src/filedata.h
@@ -330,6 +330,22 @@ class FileData::FileList
 	static void recursive_append_full(GList **list, GList *dirs, SortType method, gboolean ascend, gboolean case_sensitive);
 };
 
+/**
+ * @class FileDataRef
+ * @brief RAII object that holds a ref to the specified FileData.
+ */
+class FileDataRef
+{
+    public:
+	explicit FileDataRef(FileData &fd);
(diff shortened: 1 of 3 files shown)


xsdg:
@qarkai please review when you have a chance

qarkai (approved):
LGTM!

caclark:
Implemented in commit 1464215fcf58dfaa4d43985ce728f8585e1279b6

PR #1421, opened 2024-07-02
Deduplicate widget_set_cursor()

qarkai:
Use it to deduplicate more code.


caclark:
Implemented in commit 768b645b02608d48e90fca8fee833855064f4536

PR #1422, opened 2024-07-03
Use std::to_string() to avoid manual memory management

qarkai:


caclark:
Implemented in commit 6ecce328f5e53ea38d3678ac48c21ff65047c3e3

PR #1423, opened 2024-07-04
Encapsulates FileData globals into a dependency-injectable structure

xsdg:
This both allows us to potentially use getters and setters down the line (which could be useful if we need to make access to this object thread-safe), and also makes it easy to write unit tests that don't interfere with each other (which is important if, for instance, one test has an assertion failure and returns before it's able to clean up the common state in an appropriate way).

### DIFF
--- a/src/filedata.h
+++ b/src/filedata.h
@@ -22,6 +22,8 @@
 #ifndef FILEDATA_H
 #define FILEDATA_H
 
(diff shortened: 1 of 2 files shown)


caclark:
Implemented in commit a8fa728ff63043c170b4cac017f995b259f4f4fb

PR #1425, opened 2024-07-04
Fix memory leaks after read_char_option() calls

qarkai:
Use `g_autofree` macros.
Also remove redundant `g_strdup()`.


caclark:
Implemented in commit 8aa9144dc4c5a9982cdb0700a1a36b685c1e7fcd

PR #1426, opened 2024-07-05
Fully encapsulates `FileData` sorting behavior, adds unit tests, and fixes bugs.

xsdg:
This gets rid of the last `FileData` globals, so that the sorting settings are properly-encapsulated again.  With that done, adds unit tests (which found a number of bugs, both independently and in tandem with valgrind), and fixes the bugs.

Also, makes `FileDataRef` more generally useful.

Finishes milestone 5 from #1369

### DIFF
--- a/src/filedata.cc
+++ b/src/filedata.cc
@@ -148,9 +148,9 @@ void file_data_disable_grouping_list(GList *fd_list, gboolean disable)
 }
 
 
-gint filelist_sort_compare_filedata(const FileData *fa, const FileData *fb)
+gint filelist_sort_compare_filedata(const FileData *fa, const FileData *fb, FileData::FileList::SortSettings *settings)
 {
-	return FileData::FileList::sort_compare_filedata(fa, fb);
+	return FileData::FileList::sort_compare_filedata(fa, fb, settings);
 }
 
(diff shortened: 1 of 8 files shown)


caclark:
Implemented in commits  d9e680c6e3053dc1af3cdd740492e27b91d9823b e9cf6893ce70fe755d55e0f1314759d2ffc3c12a b1e6f405dc5db3f81fff31219f31d03533d3620b d388c93dac3cf54c08298fd413d34c0a292d2eba 14bbbe87aae0d649b27614800bc5901ba890f0e6

qarkai:
> Finishes milestone 5 from #1369

Does it mean that #1369 is implemented and could be closed?

xsdg:
> > Finishes milestone 5 from #1369
> 
> Does it mean that #1369 is implemented and could be closed?

Not yet, no.  The main reason is that I only predicted this would be the last step, but I got to it earlier than expected.  But also, I didn't realize that there were extra globals that weren't at the top of the file, which means milestone 5 isn't actually complete yet:
- https://github.com/BestImageViewer/geeqie/blob/8cdab9999d427756f8e0ab3616f854938796ff5c/src/filedata/filedata.cc#L1324
- https://github.com/BestImageViewer/geeqie/blob/8cdab9999d427756f8e0ab3616f854938796ff5c/src/filedata/filedata.cc#L2789

PR #1427, opened 2024-07-07
Fix markdownlint warnings

qarkai:
Add alt text to Geeqie icon.
Remove obsolete `#requirements` link.
Also use code blocks instead of raw text for shell commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Package Status:
 [![Packaging status](https://repology.org/badge/tiny-repos/geeqie.svg)](https://repology.org/project/geeqie/versions)
 [![latest packaged version(s)](https://repology.org/badge/latest-versions/geeqie.svg)](https://repology.org/project/geeqie/versions)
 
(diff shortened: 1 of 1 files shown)


caclark:
> Add alt text to Geeqie icon, also use link to local file.

I intentionally included the Geeqie icon as base64 text so that the icon shows if the README file is downloaded and viewed, even if  Geeqie is not installed.

The icon is also displayed if the user selects README from the Geeqie Help menu.

I know that the icon does not show on the GitHub page (I do not know why), but the Geeqie icon is shown elsewhere on that page.

Is it possible to have the alt-text and keep the icon within the file?

Are there any other people with for-or-against for this topic?

qarkai:
I'll revert link to local file.

qarkai:
Reverted.

caclark:
Implemented in commit c106e4d7a5098ecd0ea4fcfd8eaf55cca6de0a70

PR #1429, opened 2024-07-08
Fix memory leak in gq_gtk_icon_theme_load_icon_copy()

qarkai:


caclark:
Implemented in commit 7d6eb69ddbb0ed6135878d22df7501ec599604d5

PR #1430, opened 2024-07-09
Handle seg.faults during test runs

caclark:
@xsdg 

I would like to use G_DEBUG=fatal warnings during test runs to force run-time warnings to be seg. faults.

The problem is to differentiate between Geeqie exiting normally and seg. faulting during the shutdown code.

[I tried pipes, bash coproc, wrapper scripts without success]

This proposed solution removes some of your code and relies on the "nuclear" option of the Meson test timeout to kill the process if the problem is not handled within the script.

Do you have another solution?

### DIFF
--- a/src/bar-gps.cc
+++ b/src/bar-gps.cc
@@ -797,7 +797,6 @@ static GtkWidget *bar_pane_gps_menu(PaneGPSData *pgd)
 	GtkWidget *map_centre;
 	ChamplainMapSourceFactory *map_factory;
 	GSList *map_list;
-	ChamplainMapSourceDesc *map_desc;
 	const gchar *current;
 
 	menu = popup_menu_short_lived();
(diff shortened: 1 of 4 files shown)


xsdg:
> I would like to use G_DEBUG=fatal warnings during test runs to force run-time warnings to be seg. faults.
> 
> The problem is to differentiate between Geeqie exiting normally and seg. faulting during the shutdown code.
> 
> [I tried pipes, bash coproc, wrapper scripts without success]
> 
> This proposed solution removes some of your code and relies on the "nuclear" option of the Meson test timeout to kill the process if the problem is not handled within the script.
> 
> Do you have another solution?

I think this is the point where a different language might be more appropriate.  Namely, it seems like the goal is to do these three things at once:
1) Get geeqie's exit code
2) Get geeqie's text output
3) Do both 1 and 2 asynchronously…
4) …without waiting indefinitely

And as far as I can tell, that combination isn't possible in bash, but is pretty trivial in a language like Python.  I'll whip something up, and you can decide whether the switch is worthwhile.  My personal sense is that it's good to keep test execution times short (including failures), and using bash for this is already making things a bit circuitous

PR #1431, opened 2024-07-09
Fix g_slist_free() memory leaks

qarkai:
Simplify a bit.


caclark:
Implemented in commit 3b049d340163373759cd525c4fa17c807a330fe5

PR #1433, opened 2024-07-10
Add member functions to ActionItem to encapsulate memory management

qarkai:
Convert return value of `get_action_items()` to `std::vector<ActionItem>`.
Convert `action_item_compare_label()` to `ActionItem::has_label()`.
Remove `action_items_free()`.
Simplify related code.

PR #1434, opened 2024-07-11
Use bar_pane_gps_get_map_id() to avoid duplication

qarkai:
Use `g_autoptr` to simplify `bar_pane_gps_get_map_id()`.


qarkai:
Checks should be fine after adding commit from #1433 to master.

caclark:
Implemented in commit 66916334417809f214b0f00a2db1b1fc8cef84a0

PR #1435, opened 2024-07-12
Simplify GObject data destroying

qarkai:
Common pattern to associate to GObject and later free data is:

```c++
g_object_set_data(object, "key", data);
g_signal_connect(object, "destroy", G_CALLBACK(destroy_cb), data);
```

where `destroy_cb` usually defined as

```c++
void destroy_cb(GtkWidget *, gpointer data)
{
	// free data
}
```

Connecting to `"destroy"` signal is redundant since the same could be achieved using
```c++
g_object_set_data_full(object, "key", data, destroy_cb);
```

Also unused `GtkWidget *` parameter is not required anymore.


caclark:
Implemented in commit a9b1a470214e7dd739af206f9d3882829171d902

PR #1436, opened 2024-07-14
Fix Exiv2::enableBMFF() deprecation warning for exiv2 version >= 0.28.3

qarkai:


caclark:
Implemented in commit 13c1fcd912c0a89712faae45e4544706384485ce

PR #1438, opened 2024-07-17
Move duplicated code to bookmark_icon()

qarkai:


qarkai:
Implemented in commit 4b5e82f.

PR #1439, opened 2024-07-20
Use std::array for GtkTargetEntry arrays

qarkai:
Drop separate count variables.
Move to anonymous namespace.
Also convert some enums to constants.


caclark:
Implemented in commit 1bc784e41452b991722a09df39e09299d2152060

PR #1440, opened 2024-07-20
Convert string_list_free() to GDestroyNotify type

qarkai:
Move it to file where it's actually used.
Replace other calls with direct `g_list_free_full()`.


caclark:
Implemented in commit ade0ac9a8d6e4ae95a85ab0e41b01eef9143dcc3

PR #1441, opened 2024-07-21
Deduplicate *_path_to_clipboard()

qarkai:


src/utilops.cc
```diff
@@ -3286,50 +3285,31 @@ void file_util_copy_path_to_clipboard(FileData *fd, gboolean quoted, gboolean ac
...
+		GList *path_list = nullptr;
 
-		while (work)
+		for (GList *work = fd_list; work; work = work->next)
 			{
-			fd = static_cast<FileData *>(work->data);
-			work = work->next;
+			auto *fd = static_cast<FileData *>(work->data);
 
 			if (!fd || !*fd->path) continue;
```
qarkai, line 3297: @xsdg Does it make sense to add this condition to `FileList::to_path_list()`?

xsdg: Probably not without checking the assumptions from the callsites.  In particular, I can imagine a caller assuming that the returned list will have the same length as the list that was passed in, which could cause some subtle bugs if that is violated.

On the upside, the function is only called from two places (`bar-sort.cc` and `uri-utils.cc`).  Unfortunately, the usage in `uri-utils.cc` (in `uris_from_filelist`) is used in a bunch of different places, and seems to be related to selection logic.

caclark:
Implemented in commit 21246abe355286e95dfdbe4298b416913b570bc9

PR #1442, opened 2024-07-21
Drop duplicated drag-n-drop targets enum

qarkai:


caclark:
Implemented in commit 7b2461bf935053b82a2b3427ccd587adde6fc0b4

PR #1443, opened 2024-07-22
Fix memory leaks after g_object_get() calls

qarkai:


caclark:
Implemented in commit 29a277a85419ad8854a4a238a5554873cc1f3bef

PR #1444, opened 2024-07-24
Simplify cache maintenance a bit

qarkai:
Replace `extension_find_dot()` with `strrchr()`.
Deduplicate `cache_maint_removed()` and inline `cache_file_remove()`.
Deduplicate `cache_maint_moved()` and inline `cache_file_move()`.
Flatten `cache_maint_copied()`.
Unify names of variables.


caclark:
Implemented in commit 8cdab9999d427756f8e0ab3616f854938796ff5c

PR #1445, opened 2024-07-25
Add wrappers for cache_get_location()

qarkai:
`cache_get_location()` is called for two reasons:
1) get cache directory and immediately create it.
2) get cache filename.

Add `cache_create_location()` for (1).
Add `cache_get_location()` wrapper for (2). This allows to remove `mode_t` from parameters.
Move old `cache_get_location()` implementation to anonymous namespace.
Fix IWYU warnings after removing dependency on `mode_t`.


caclark:
Implemented in commit 998829abeeaa322eabd1890ff17c375cbaa4185e

PR #1447, opened 2024-07-27
Simplify static local variables initialization

qarkai:


caclark:
Implemented in commit 8894044c2d38a8088d843d28ff96918c5913a7e0

PR #1448, opened 2024-07-29
Fix Metadata table columns sort

qarkai:
Simplify `advanced_exif_sort_cb()`.


caclark:
Implemented in commit 3350715cd884b998f35a9a56528613f4c3935951

PR #1451, opened 2024-07-30
Avoid logging nullptr path in cache_maint_moved()

qarkai:
Move logging to `cache_create_location()`.
Simplify `cache_maint_moved()`.


caclark:
Implemented in commit a8447ba320c340dc8c5096c9050dd267d7e41923

PR #1452, opened 2024-07-31
Convert print_pref_store() to print_get_page_text() to avoid duplication

qarkai:


caclark:
Implemented in commit 829500ea3d5ec86573b5d7f3dcf1ff153c9a095f

PR #1454, opened 2024-08-01
Deduplicate print layout creation in draw_page()

qarkai:


caclark:
Implemented in commit 10b5d17425ade6ae64c6166a56be35f212d66778

PR #1455, opened 2024-08-02
Deduplicate text positions callbacks in print dialog

qarkai:


caclark:
Implemented in commit b259d73a75baa094680537392bad4228a8a1ebb0

PR #1457, opened 2024-08-03
Use TextPosition type for image_text_position and page_text_position

qarkai:


caclark:
Implemented in commit 915e28b75d5921c319fa3f452698f6f0566532e0

PR #1459, opened 2024-08-04
Simplify orientation handling a bit

qarkai:


caclark:
Implemented in commit 06d94efeb48385c2988ebde1edaf498e28e0f452

PR #1460, opened 2024-08-04
Simplify adding layouts in draw page print callback

qarkai:


caclark:
Implemented in commit 3f4fb64a98b9a9541081ca7fe7e225863fd42468

PR #1461, opened 2024-08-04
Fix memory leaks after g_list_free() calls

qarkai:


caclark:
Implemented in commit 4922f3d35cfedc86b6ce4b93369dbf6522ff8e33

PR #1463, opened 2024-08-05
Add ClipboardAction enum instead of action_copy flag

qarkai:
See https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#i4-make-interfaces-precisely-and-strongly-typed and
https://alexkondov.com/should-you-pass-boolean-to-functions/#context for rationale.


caclark:
Implemented in commit 49230f5f22341c041d290257ac3e64d5472129e9

PR #1464, opened 2024-08-05
Extract duplicated code to cache_maintain_data_new()

qarkai:
Also make `cache_maintain_home()` internal.


caclark:
Implemented in commit 1944cd872c0dec6271d049b801bae2aedd4b9e26

PR #1466, opened 2024-08-06
Simplify log_domain_print_message()

qarkai:
Use `g_regex_match_simple()` instead of POSIX regex functions.
Make static.
Make `buf` parameter `const`, move freeing `buf` to callers.


caclark:
Implemented in commit 03cf3a5e6206f3ac7ca4cbd2ebd0e130675076a3

PR #1467, opened 2024-08-07
Fix memory leaks found by PVS-Studio static analyzer

qarkai:
Simplify code a bit.


caclark:
Implemented in commit e6ee3fdbfc958635563732e471c3a79191b5b9cf

PR #1470, opened 2024-08-09
Fix couple memory leaks

qarkai:


caclark:
Implemented in commit 72f9727b234d421256dd107ba157a93363c19b23

PR #1471, opened 2024-08-09
Declare internal functions static

qarkai:
Move whole block of static functions to anonymous namespace.
Remove found unused `exif_add_item()` function.


caclark:
Implemented in commit ea92dab35a3763f91547ce662c3ad42f4e2c350b

PR #1473, opened 2024-08-12
Fix country_name -> country_code misprint in gr_file_info()

qarkai:
Reduce scope of some variables to avoid misprint warnings.


caclark:
Implemented in commit 5029e4fdf753269661f5d0e98702073855f342a2

PR #1474, opened 2024-08-13
Simplify image_get_path() usage

qarkai:
Drop odd const casting.
Cache result in variable.
Drop unnecessary string duplication.


caclark:
Implemented in commit 308e1efe0cd405323497da06f9fccf4b8f935e31

PR #1475, opened 2024-08-14
Fix memory leaks after pixbuf_inline() calls

qarkai:
Also replace GtkAboutDialog's logo property with logo-icon-name to avoid manual creation of pixbuf.


caclark:
Implemented in commit 1c1430a8159dc29d83b5b25dccf43e68617a9fc0

PR #1476, opened 2024-08-15
Encapsulate GQParserData related types and functions and simplify a bit

qarkai:
Use `std::stack` for `GQParserData::parse_func_stack`.
Create `parser_data` on stack.
Convert `parser` to static local variable.


caclark:
Implemented in commit 1d4259cf4e7b9962b7b387d1cedbb29108e32560

Issue #1307, opened 2024-04-02
Remote --cache-maintenance causes out-of-memory


caclark:
**Setup (please complete the following information):**
 - Distribution: Ubuntu
 - Distribution release: 23:10
 - Geeqie version: All
 
**Describe the bug**

Running `geeqie --cache-maintenance` on a directory of 135, mostly raw, files causes out-of-memory on  a computer with 4GB of memory.

qarkai:
@caclark Could you check on your files after dd16d83?

caclark:
@qarkai 
Thanks.

Fixed in commit dd16d83

PR #1477, opened 2024-08-16
Fix memory leaks in cache maintenance

qarkai:
Found by Valgrind.
Related to #1307.


caclark:
Implemented in commit 94392d0b998af790456a7d4adb8150156e823c69

PR #1479, opened 2024-08-17
Generate list of authors respecting .mailmap

qarkai:
Reduces duplication a bit.


caclark:
Implemented in commit b50da2636d2fccc7944530cf2c21ef8c75160d12

PR #1480, opened 2024-08-17
Fix printf format strings

qarkai:
Fix `FileData::text_from_size()`, add more values to test.
Also check if `%s` argument could be `nullptr`.


caclark:
Implemented in commit e95256c4e7656c2b99d8754386421675b8b0e6be

PR #1482, opened 2024-08-18
Remove redundant code

qarkai:
Duplicated and always true/false conditions, unnecessary assignments, etc.


caclark:
Implemented in commit 229fda4559805ba38a4e6b32ac72c347a87188f3

PR #1483, opened 2024-08-18
Optimize a bit

qarkai:
Use prefix increment operator for iterators.
Search for character instead of one-character string.
Avoid extra `strdup()`.


caclark:
Implemented in commit aab1512aa142536cd91c2ba990773c5021e3be09

PR #1485, opened 2024-08-19
Use icon-name in GtkAboutDialog

qarkai:
Also simplify initialization of `artists`.


caclark:
Implemented in commit c0bb6fb3ccf2b295670bfca5f051894e43601c70

PR #1488, opened 2024-08-20
Simplify ImageLoaderJ2K::write()

qarkai:
Fix memory leaks of `opj_*_t` structs.
Check `opj_codec_set_threads()` return value.
Rename `opj_buffer_info_t` to `OpjBufferInfo`.
Allocate `decode_buffer` on stack.
Reduce scope of variables.

### DIFF
--- a/src/image-load-j2k.cc
+++ b/src/image-load-j2k.cc
@@ -36,6 +36,10 @@
 namespace
 {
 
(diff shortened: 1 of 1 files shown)


caclark:
Implemented in commit 45f84b28b5ebca520850cbb78e29641c2b81a6cd

PR #1489, opened 2024-08-21
Remove unused opj_write_to_buffer()

qarkai:
Since `input` parameter in `opj_stream_create_buffer_stream()` is always `OPJ_TRUE`.
`OpjBufferInfo` members now readonly and could be const.
Remove now redundant `buf_copy`.
Also inline `opj_stream_create_buffer_stream()` and simplify code.


caclark:
Implemented in commit 7f5a22d8b6c2f7930547dd69339d3968af4d171b

PR #1490, opened 2024-08-23
Remove ImageLoader dependency from image-load-libraw

qarkai:
Also change type of `data_len` parameter in `libraw_get_preview()`.


caclark:
Implemented in commit d5789f5217ab93235f1d8f3ff58a776a4a0372c3

PR #1491, opened 2024-08-24
Fix memory leaks in ImageLoaderDJVU::write()

qarkai:


caclark:
Implemented in commit 6ef56ba69e9c863ae1a3f3d3b67d0617fe905e7a

PR #1492, opened 2024-08-24
Deduplicate file mapping

qarkai:


caclark:
Implemented in commit 437b050192b25c346253482a03e3ffd13a836320

PR #1493, opened 2024-08-25
Fix memory leaks in loading raw files

qarkai:
Found by Valgrind.
Related to #1307.


caclark:
Implemented in commit dd16d83aba5bbbb1a4ca1628f753b5c692f77d1b

PR #1494, opened 2024-08-25
Fix memory leak in loading psd files

qarkai:
Found by Valgrind.
Related to #1307.


caclark:
Implemented in commit ec8abe313df02222d8f3bc69b08d79a7a5bcd5fd

PR #1495, opened 2024-08-25
Add is_jpeg_container() to avoid using magic numbers

qarkai:


caclark:
Implemented in commit 22512962558e44ed1812ff0c4ea715a5659f39c3

PR #1496, opened 2024-08-25
Simplify file_in_path()

qarkai:
Use `g_strsplit()` instead of manual parsing.


caclark:
Implemented in commit b5d07cae7800f053809ff0fb0955a030cd37ff73

PR #1497, opened 2024-08-27
Use LibRaw to check if thumb is jpeg

qarkai:
Also fix potential memory leak of mapped file.


caclark:
Implemented in commit 01865a69d909253012dba99f01c9125110d641c0

PR #1498, opened 2024-08-27
Simplify download_web_file()

qarkai:
Flatten code and fix couple memory leaks.


caclark:
Implemented in commit cc0ca5b924dfe2bf3d749caa895fcfddcea02abc

PR #1500, opened 2024-08-29
Use strptime() for datetime parsing

qarkai:
Simplify related code a bit.


caclark:
Implemented in commit 39e929b2df4dbc3ce385dc0757deb5685e72e9c6

PR #1502, opened 2024-08-30
Pass cheap-to-copy arguments by copy

qarkai:
`GdkPoint`, `GdkRectangle`, `PanColor` sizes are not exceed 16 bytes.
See https://pvs-studio.com/en/docs/warnings/v835/ for the rationale.


caclark:
Implemented in commit 1a356cbb6d4923c5e7cee044958c6a58eb701208

PR #1504, opened 2024-08-31
Use GdkRectangle in DialogWindow

qarkai:


caclark:
Implemented in commit 080589d3ee834f7669b635cdb312f76841bbdfa1

PR #1506, opened 2024-09-01
Add LayoutOptions::SortParams to reduce code duplication

qarkai:


caclark:
Implemented in commit 5c0affe5fccae9c04960bf5d04c1bde262d75fd7

Issue #522, opened 2017-09-26
geeqie: saves state information under ~/.config


caclark:
This is from:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=623993

> geeqie saves the names of opened directories into
~/.config/geeqie/history.  This is wrong because only configuration
files should be under .config.  The history file should be under
/.local/share/geeqie/ , which is already in use.

Maybe Geeqie .desktop files are in the wrong place also.

qarkai:
> Maybe Geeqie .desktop files are in the wrong place also.

I don't see .desktop files in `~/.config/geeqie`.

Is there any other data that should not be in `~/.config/geeqie` directory?

caclark:
> I don't see .desktop files in ~/.config/geeqie.

Seven years is a long time to remember things.... but I suspect it is this:

If you open `Edit/Configure Plugins` , select an entry and then edit it, when it is saved it is stored in `$HOME/.config/geeqie/applications`.

The correct location is `XDG_DATA_HOME`, as specified in `https://specifications.freedesktop.org/basedir-spec/latest/index.html`

(But right at this moment I do not find the paragraph in freedesktop.org that specifically says that  `XDG_DATA_HOME` should be used e.g. is a desktop file a configuration file or a data file)

caclark:
> Is there any other data that should not be in ~/.config/geeqie directory?

I have not found a clear, unambiguous, definition for what should be in `XDG_DATA_HOME` and what should be in `XDG_CONFIG_HOME`.

The problem I see is that if  "data" is moved from one location to another, on the first run after the code change the program has to be able to transfer the user's data from one place to another so that nothing is lost.

PR #1510, opened 2024-09-02
Use XDG data directory for history file

qarkai:
Also use `g_get_user_*_dir()` functions instead of `xdg_dir_get()` implementation.
Related to #522.


caclark:
Implemented in commit a3abbc653839da413bc5a168457c6472872230a2

PR #1512, opened 2024-09-04
Drop redundant comparison

qarkai:


caclark:
Implemented in commit be83ce6806033c1f979608b9a5cac175b8c1a94f

PR #1514, opened 2024-09-07
Check wkhtmltopdf dependency for help.pdf build

qarkai:
Note: wkhtmltopdf is archived and no longer maintained.


caclark:
This was a very bad idea to use wkhtmltopdf (although it seemed quick-and-easy-and-no-thinking at the time - and the implementation was bad).

I am going to revert that commit.

Deriving the pdf from the xml is a better way.

qarkai:
Close then.

PR #1516, opened 2024-09-08
Use EXIT_SUCCESS and EXIT_FAILURE macros

qarkai:


caclark:
Implemented in commit 574f85a7ccd049f9b87eb50133cb9cd5a6cf4afd

PR #1518, opened 2024-09-08
Fix memory leak in bar_pane_keywords_destroy()

qarkai:


caclark:
Implemented in commit 6dcdf60896eff4931db71936c28365c8182c622c

Issue #1520, opened 2024-09-13
Command-line utility for duplicate image finding and processing


porridge:
**Setup (please complete the following information):**
 - Geeqie version [geeqie --version]: irrelevanat

**What is your question**

Would you be open to shipping an additional binary that reuses the code in `similar.cc` for comparing images, but is meant for automated or semi-automated workflows to find and eliminate duplicate images?

**Additional context**

I find the duplicate image logic in geeqie to be of excellent quality. 🤩 

However the GUI interface is not convenient/flexible enough for my use case. I was looking for a tool such as [findimagedupes](https://github.com/jhnc/findimagedupes) which makes it possible to plug in code that can decide which files in a set to delete, however I found its comparison logic to be [less than stellar](https://github.com/jhnc/findimagedupes/issues/21).

After dusting off my C++ knowledge (and commenting out a few lines in `similar.cc`), I managed to come up with an alternative frontend to your code, that - although a bit naive/inefficient - works exactly as I imagined and started saving me literally hours of time per year.

This code currently weighs 155 lines of C++, and 50 or so lines of example Python code to automate the decision of whether to delete files or ask for confirmation.

I would like to share it with others, and I can either:
1. start my own project and copy `similar.*` with minimal modifications over,
2. ask you to add it to geeqie 😊 

I'm asking this preliminary question first, rather than send a PR straight away, because I would need to spend some more time to properly do option processing before sending a well-formed PR. I'm happy to take a stab at this, but would rather avoid doing so if the answer is "no" anyway 😄

caclark:
From my point view, adding extra command line options is not a problem. If people are going to use such a feature it would be better for them to do it via Geeqie rather than via another project.

Although, if implemented, there will be a configuration option to disable it, I am not sure about the implications of making python a dependency. Perhaps @gusnan has an opinion.

The only python dependency in the project so far is involved with running tests, so is not part of the distribution.

porridge:
Sorry, I should have pointed out that the Python part is completely optional. Basically how the main program works is that it runs some command once for each group of duplicates (passing them as arguments). By default the command is just `echo` but you can specify anything you like.

I just happened to use Python to create a script that accepts the filenames as parameters, and if their names match certain patterns (in my case this hints on their source) deletes smaller files without asking for confirmation.

porridge:
[...]

PR #1524, opened 2024-09-28
Add a remote command for batch duplicate finding.

porridge:
Here is my first stab at it. I know I at least need to adjust for the accepted coding style, but please let me know if there are any other major things that need changing.

Based on https://github.com/porridge/image-duplicate-finder

Closes: #1520


src/pic.h
```diff
@@ -0,0 +1,42 @@
...
+ */  
+
+#include <set>
+#include <string>
+#include <gdk-pixbuf/gdk-pixbuf.h>
+#include <glib/gtypes.h>
+
+#include "similar.h"
+
+class pic {
```
xsdg, line 32: Please at least add some minimal documentation, in doxygen format.

That should include why this class is called "pic", what the class usage pattern should be, and what the method arguments mean.

See examples:
- https://github.com/BestImageViewer/geeqie/blob/8cdab9999d427756f8e0ab3616f854938796ff5c/src/filedata.h#L396
- https://github.com/BestImageViewer/geeqie/blob/8cdab9999d427756f8e0ab3616f854938796ff5c/src/pixbuf-util.cc#L1581

src/remote.cc
```diff
@@ -682,6 +688,102 @@ static void gr_slideshow_delay(const gchar *text, GIOChannel *, gpointer)
...
+static void gr_process_duplicates(const gchar *, GIOChannel *, gpointer data)
+{
+	auto remote_data = static_cast<RemoteData *>(data);
+    std::map<std::string, std::unique_ptr<pic>> pics;
+	GList *work = remote_data->file_list;
+	while (work)
+	{
+		FileData *fd = static_cast<FileData *>(work->data);
+        std::string name(fd->path);
+        pics[name] = std::unique_ptr<pic>(new pic(fd->path));
```
xsdg, line 732: You should generally be using `std::make_unique` to create `unique_ptr`s.  See the example section in
https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique

For context when referring to language docs, Geeqie is using C++14 (as can be seen in the `meson.build` project section)

src/remote.cc
```diff
@@ -682,6 +688,102 @@ static void gr_slideshow_delay(const gchar *text, GIOChannel *, gpointer)
...
+		work = work->next;
+	}
+	DEBUG_1("processing %d files in set", pics.size());
+
+	// Compute similarity score for every pair, build equivalence sets.
+    for (auto a = pics.begin(); a != pics.end(); ++a) {
```
xsdg, line 738: Please use meaningful variable names for temporaries.  Maybe `left_iter` and `right_iter`?  `left_pic` and `right_pic`?  Doesn't have to be those, but the meanings should be self-evident from the name.

src/remote.cc
```diff
@@ -682,6 +688,102 @@ static void gr_slideshow_delay(const gchar *text, GIOChannel *, gpointer)
...
+        auto b = a;
+        b++;
+        for (; b != pics.end(); ++b) {
+            double similarity = a->second->compare(*b->second);
+            DEBUG_1("%s vs %s: %f", a->second->name.c_str(), b->second->name.c_str(), similarity);
+            if (similarity < options->duplicates_similarity_threshold)
+                continue;
+            a->second->equivalent.insert(b->second->equivalent.begin(), b->second->equivalent.end());
+            for (auto const &f: a->second->equivalent) {
```
xsdg, line 747: Meaningful variable name

src/remote.cc
```diff
@@ -682,6 +688,102 @@ static void gr_slideshow_delay(const gchar *text, GIOChannel *, gpointer)
...
+                pics[f]->equivalent.insert(a->second->equivalent.begin(), a->second->equivalent.end());
+            }
+        }
+    }
+
+    std::set<std::string> printed;
+    for (auto const &f: pics) {
```
xsdg, line 754: Meaningful variable name

src/remote.cc
```diff
@@ -682,6 +688,102 @@ static void gr_slideshow_delay(const gchar *text, GIOChannel *, gpointer)
...
+        if (f.second->equivalent.size() < 2)
+            // skip this pic if not similar to any other one but itself
+            continue;
+        if (printed.find(f.second->name) != printed.end())
+            // skip this pic if it was already printed (when processing a similar image)
+            continue;
+        std::vector<const char *> cmd;
+		cmd.push_back(options->duplicates_program);
+        for (auto const &e: f.second->equivalent) {
```
xsdg, line 763: Meaningful variable name

src/remote.cc
```diff
@@ -682,6 +688,102 @@ static void gr_slideshow_delay(const gchar *text, GIOChannel *, gpointer)
...
+            cmd.push_back(e.c_str());
+            printed.insert(e);
+        }
+        cmd.push_back(NULL);
+        pid_t pid = fork();
+        if (pid == -1) {
+            perror("fork");
+            exit(1);
```
xsdg, line 771: `exit()` is not called directly from anywhere else in this file, and geeqie does not generally use `perror`.  The best course of action here is likely to log an error and return.  See an example here: \
https://github.com/BestImageViewer/geeqie/blob/8cdab9999d427756f8e0ab3616f854938796ff5c/src/remote.cc#L541

With that said, it's probably a lot safer to use an API like https://docs.gtk.org/gio/ctor.Subprocess.new.html instead of `fork`ing directly.  That said, @qarkai is our resident expert on glib APIs, so I would defer to him.

src/pic.h
```diff
@@ -0,0 +1,42 @@
...
+public:
+    pic(char const *cname);
```
xsdg, line 34: This class violates the rule of 3/5/0.  See: \
https://en.cppreference.com/w/cpp/language/rule_of_three

src/pic.cc
```diff
@@ -0,0 +1,53 @@
...
+
+pic::pic(char const *cname): name(cname), equivalent{name}, sim(NULL) {
+    GError *err = NULL;
+    GdkPixbuf *buf = gdk_pixbuf_new_from_file(cname, &err);
+    if (buf == NULL) {
+        fprintf(stderr, "Unable to read file %s: %s\n", cname, err->message);
+        g_error_free(err);
+        return;
+    }
+    sim = image_sim_new_from_pixbuf(buf);
```
qarkai, line 35: Consider moving `sim` initialization to static method or anonymous function.

src/pic.cc
```diff
@@ -0,0 +1,53 @@
...
+    g_object_unref(buf);
```
qarkai, line 36: Consider using `g_auto*` macros instead of manual memory management.

src/pic.cc
```diff
@@ -0,0 +1,53 @@
...
+ *
+ *
+ * Helper class for computing equivalence sets of pictures.
+ *
+ */  
+
+#include "pic.h"
+
+pic::pic(char const *cname): name(cname), equivalent{name}, sim(NULL) {
+    GError *err = NULL;
```
qarkai, line 28: In geeqie this variable is almost always called `error`.

src/remote.cc
```diff
@@ -682,6 +688,102 @@ static void gr_slideshow_delay(const gchar *text, GIOChannel *, gpointer)
...
+			printf_term(TRUE, "Image similarity threshold out of range (%d to %d)\n", 0, 100);
+			return;
+			}
+		}
+	else
+		{
+		n = 99;
+		}
+
+	options->duplicates_similarity_threshold = static_cast<guint>(n);
```
qarkai, line 712: Looks like `n` is redundant. If you'd like to have intermediate variable consider changing its type to `guint`.

src/remote.cc
```diff
@@ -682,6 +688,102 @@ static void gr_slideshow_delay(const gchar *text, GIOChannel *, gpointer)
...
+	gint n;
+	gint res;
+
+	res = sscanf(text, "%d", &thresh);
+	if (res == 1)
+		{
+		n = thresh;
+		if (n < 0 || n > 100)
+			{
+			printf_term(TRUE, "Image similarity threshold out of range (%d to %d)\n", 0, 100);
```
qarkai, line 703: Range limits are duplicated on line 701. Consider moving them to constants.
Probably would be useful to print `thresh` value here.

src/remote.cc
```diff
@@ -682,6 +688,102 @@ static void gr_slideshow_delay(const gchar *text, GIOChannel *, gpointer)
...
+		n = 99;
+		}
+
+	options->duplicates_similarity_threshold = static_cast<guint>(n);
+	DEBUG_0("threshold set to %d", options->duplicates_similarity_threshold);
+}
+
+static void gr_duplicates_program(const gchar *text, GIOChannel *, gpointer)
+{
+	g_strdup(options->duplicates_program);
```
qarkai, line 718:
```suggestion
	g_free(options->duplicates_program);
```

src/remote.cc
```diff
@@ -682,6 +688,102 @@ static void gr_slideshow_delay(const gchar *text, GIOChannel *, gpointer)
...
+	options->duplicates_program = g_strdup(text);
+	DEBUG_0("duplicates program set to \"%s\"", options->duplicates_program);
+}
+
+static void gr_process_duplicates(const gchar *, GIOChannel *, gpointer data)
+{
+	auto remote_data = static_cast<RemoteData *>(data);
+    std::map<std::string, std::unique_ptr<pic>> pics;
```
qarkai, line 726: Please use tab for indentation.

src/remote.cc
```diff
@@ -682,6 +688,102 @@ static void gr_slideshow_delay(const gchar *text, GIOChannel *, gpointer)
...
+	GList *work = remote_data->file_list;
+	while (work)
```
qarkai, line 728: Consider using `for` loop to reduce scope of `work`.

src/remote.cc
```diff
@@ -682,6 +688,102 @@ static void gr_slideshow_delay(const gchar *text, GIOChannel *, gpointer)
...
+	{
```
qarkai, line 729: Please fix braces indentation all over PR.

src/remote.cc
qarkai: I haven't used GSubprocess much. geeqie uses `g_spawn_async_with_pipes` for opening external editor, see `editor_command_one` in `src/editors.cc:1053`.

qarkai: Geeqie uses `g_spawn_async_with_pipes()` for external editor. See `editor_command_one()` in `src/editors.cc:1053`.

caclark:
In the latest commit, the command line handling has changed. The attached .diff are the changes I think are necessary to conform with the new code:

[1524-1.diff.gz](https://github.com/user-attachments/files/17365613/1524-1.diff.gz)

caclark:
Revised .diff file.
Please note that these changes are merely suggestions.

You can run the project static tests locally before making a pull request by:
./scripts/test-all.sh

[1524-2.diff.gz](https://github.com/user-attachments/files/17375869/1524-2.diff.gz)

porridge:
Thank you for the reviews @xsdg and @qarkai and the rebase @caclark!

Here's a new iteration, where I also removed the need to run in "remote" mode.

I think I addressed all the comments, apart from one, about using `fork`. As far as I could tell, the `g_spawn_async_with_pipes` function has two problems with my use case:
1. according to its own documentation, it does not handle whitespaces in arguments well, and this is critical for reliable passing of file paths,
2. it is not possible to use it in a way which just inherits stdin and stdout, which is necessary when the program run on each duplicate set communicates with the user through the terminal.

Example usage:

```
$ ./build/src/geeqie -p ?.png
/home/porridge/Pulpit/coding/geeqie/2.png /home/porridge/Pulpit/coding/geeqie/3.png
$
```

and with debug output:

```
$ ./build/src/geeqie --debug 1 ?.png
duplicates program set to "echo"
processing 3 files in set
/home/porridge/Pulpit/coding/geeqie/1.png vs /home/porridge/Pulpit/coding/geeqie/2.png: 91,749387
/home/porridge/Pulpit/coding/geeqie/1.png vs /home/porridge/Pulpit/coding/geeqie/3.png: 91,701134
/home/porridge/Pulpit/coding/geeqie/2.png vs /home/porridge/Pulpit/coding/geeqie/3.png: 99,773284
/home/porridge/Pulpit/coding/geeqie/2.png /home/porridge/Pulpit/coding/geeqie/3.png
```

caclark:
@porridge 

I suggest you consider changing option `--process-duplicates` to `--duplicates-process`. This will group the three new commands together in the help output, and will be more logical when using command line completion. The short form option may be a bit illogical, but I do not think that matters.

There should be entries for the three command line long options in `./auto-complete/geeqie`. This is for bash command line completion.

If you wish the `--duplicates-program` to be remembered between sessions, there should be entries at about lines 409 and 915  of `./src/rcfile.cc`.

I have some other comments but it will be a few days before I write them down.

caclark:
@porridge 
I am sorry that it has taken me so long to think over this feature

@ anyone else
For comment

I think there may be another way to solve this problem. It would involve some new features that may or may not be possible.

The adding of files to the Dupes Window is a bit messy. Either by drag-and-drop or right click on files or directory - but this does not work as I expect at the moment (the right-click feature opens a new Dupes Window each time, which I do not think is correct).

1. It may be possible to include a right-click add files option to the Dupes Window

2. It may be possible to add files to the Dupes Window via the command line  e.g.
`geeqie --dupes-window-add <list of files>`
followed by the Dupes Window being opened automatically if not already open.

When the Dupes Window is open the user has the possibility to change the comparison type, the image rotation mode and other options.
When the dupes check is completed, the user has the choice of which data to select.

The selected data may be Exported to a comma-separated or tab-separated file via a right-click in the Dupes Window.

3. It may be possible to send the same selected data to the command line e.g.
`geeqie --dupes-export | cut --fields=5`

4. For the above command to be part of an automate-able sequence, it would be necessary to know when the dupes operation has finished. I have no idea at the moment.

The plugins are available via a right-click in the Dupes window. With that pretty much anything can be achieved, and might make `--dupes-export` redundant. Unfortunately plugin keyboard shortcuts are not recognized.

5. It may be possible for the Dupes Window to recognize plugin keyboard shortcuts.

6. It may be possible for there to be a `When a dupes run is finished, call this plugin` feature - but that could be dangerous for an unwary user trying to delete unwanted files.  I am not enthusiastic about this idea, but it is one way of knowing when the dupes run is completed (it would run whenever the user changes the comparison mode, for instance, and not just when the user has made the final decision).

porridge:
> @porridge
> 
> I suggest you consider changing option `--process-duplicates` to `--duplicates-process`. This will group the three new commands together in the help output, and will be more logical when using command line completion. The short form option may be a bit illogical, but I do not think that matters.

Done.

> There should be entries for the three command line long options in `./auto-complete/geeqie`. This is for bash command line completion.

I believe they are there already in the `options` variable definition? Or do you have something else in mind?

> If you wish the `--duplicates-program` to be remembered between sessions, there should be entries at about lines 409 and 915 of `./src/rcfile.cc`.

I think it's better to require the user to explicitly provide the program, in case it is destructive and the user forgot what it was set to last.

porridge:
> I think there may be another way to solve this problem. It would involve some new features that may or may not be possible.

Sounds like what you are suggesting would be more discoverable for a typical GUI user. OTOH it _might_ be less convenient for use in scripted, automated workflows.

But most importantly from my perspective - I don't feel anywhere as competent as would be required to implement what you described 😅

caclark:
@porridge 
Attached is a .diff from the current sources. I would appreciate it if you would take a look at it.
This code is not a solution - it is just a hack to demonstrate a different way of achieving this feature.

After compiling, from a terminal window run `./build/src/geeqie`
Open the dupes window and select `Compare By` to `Similarity Custom`.  Set `Custom Threshold` to a low number e.g. `50`
Close the dupes window.

From another terminal window run `./build/src/geeqie --duplicates-process <list of files.>`
For `<list of files>` just use 4 or 5 simple jpegs.

Run `./build/src/geeqie --duplicates-export`
Then try `./build/src/geeqie --duplicates-export |  tail -n +2 | cut -f 2,5`

The advantages I see are:
No additional comparison logic is required
No specific processing program is required
It is easier for users to process the data in the way they wish

If this is an acceptable solution, the `duplicates-export` function will be eliminated - the option `duplicates-process  ...` will output the required text (there is a timing problem I did not yet solve).
[1524-1.diff.gz](https://github.com/user-attachments/files/18307073/1524-1.diff.gz)

porridge:
@caclark 

> After compiling, from a terminal window run `./build/src/geeqie`
[...]
> From another terminal window run `./build/src/geeqie --duplicates-process <list of files.>` For `<list of files>` just use 4 or 5 simple jpegs.

It would be hard to incorporate this kind of usage (run a GUI-dependant geeqie session and while it's running, launch another process) into my workflow. Not impossible, but the changing window focus and juggling a process running in the background are tricky to handle.

FTR my use case is an automated pipeline that:
- moves images from input directories (to which they are first synced using syncthing from various devices) into a staging location, 
- de-duplicates them
- auto rotates and performs various other metadata fixups
- splits them into year/month/day based directory structure.

> Run `./build/src/geeqie --duplicates-export` Then try `./build/src/geeqie --duplicates-export | tail -n +2 | cut -f 2,5`

Also, simple text format makes it hard to *reliably* handle filenames which in principle can contain arbitrary whitespace characters. This could be done in a somewhat standard way using JSON format for example, but even then the need to post-process that iin order to pass further down the pipeline is yet another hurdle. This is why I found the feature of running a separate user specified command on each set of duplicates in turn particularly attractive.

> The advantages I see are: No additional comparison logic is required

Yes, that would indeed be nice. While it was super easy to reuse the algorithm for comparing two images, when I tried to reuse the logic that geeqie uses internally to process a whole list of files I got completely lost in the sources 😅

src/command-line-handling.cc
```diff
@@ -407,6 +411,114 @@ void gq_delay(GtkApplication *, GApplicationCommandLine *app_command_line, GVari
...
+		pics[name] = std::make_unique<pic_equiv>(fd);
+		}
+	DEBUG_1("processing %d files in set", pics.size());
+
+	// Compute similarity score for every pair, build equivalence sets.
+	for (auto outer_iter = pics.begin(); outer_iter != pics.end(); ++outer_iter)
+		{
+		auto inner_iter = outer_iter;
+		inner_iter++;
+		for (; inner_iter != pics.end(); ++inner_iter)
```
qarkai, line 430:
```suggestion
		for (auto inner_iter = std::next(outer_iter); inner_iter != pics.end(); ++inner_iter)
```

src/command-line-handling.cc
```diff
@@ -20,6 +20,9 @@
 
 #include "command-line-handling.h"
 
+#include <cstring>
+#include <map>
+#include <sys/wait.h>
```
qarkai, line 25: 'System' headers should be above STL headers. See `main.cc` for reference.

src/command-line-handling.cc
```diff
@@ -407,6 +411,114 @@ void gq_delay(GtkApplication *, GApplicationCommandLine *app_command_line, GVari
 	options->slideshow.delay = static_cast<gint>(n * 10.0 + 0.01);
 }
 
+void gq_duplicates_process(GtkApplication *, GApplicationCommandLine *, GVariantDict *, GList *file_list)
+{
+	std::map<std::string, std::unique_ptr<pic_equiv>> pics;
```
qarkai, line 416: `unique_ptr` seems redundant.

src/command-line-handling.cc
```diff
@@ -407,6 +411,114 @@ void gq_delay(GtkApplication *, GApplicationCommandLine *app_command_line, GVari
...
+				log_printf("subprocess failed, aborting further duplicate processing\n");
+				return;
+				}
+			}
+		}
+}
+
+void gq_duplicates_program(GtkApplication *, GApplicationCommandLine *, GVariantDict *command_line_options_dict, GList *)
+{
+	gchar *text = nullptr;
```
qarkai, line 488:
```suggestion
	const gchar *text = nullptr;
```

src/meson.build
```diff
@@ -130,6 +130,8 @@ main_sources = files('advanced-exif.cc',
 'osd.cc',
 'osd.h',
 'pan-view.h',
+'pic_equiv.cc',
```
qarkai, line 133: Please replace underscore with hyphen for consistency.

qarkai (changes requested):

caclark:
@porridge 
OK
Cache maintenance runs Geeqie in non-gui mode. I will use that as an example and try to create something better.

src/command-line-handling.cc
porridge: Thanks for the review @qarkai 
I addressed the other comments, but with my rudimentary C++ skills I cannot see how to address this one?
- replacing with `pic_equiv *` would leak memory, right?
- replacing with `pic_equiv` does not work because the assignment operator is deleted:
  ```
../src/command-line-handling.cc: In function ‘void {anonymous}::gq_duplicates_process(GtkApplication*, GApplicationCommandLine*, GVariantDict*, GList*)’:
../src/command-line-handling.cc:422:42: error: use of deleted function ‘pic_equiv& pic_equiv::operator=(const pic_equiv&)’
  422 |                 pics[name] = pic_equiv(fd);
```
  and it's deleted because of the `sim` member.

  It also seems to me like using a pointer is better than copying the sim structure around...

Comment (qarkai) [reply src/command-line-handling.cc:417]: I see. `ImageSimilarityData` lacks constructors and moving operator. Ok. let's stay with `unique_ptr` for now.

**Review by qarkai — APPROVED** [review]
LGTM.

File: src/pic-equiv.h [hunk 23]
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * @class pic_equiv
+ * @brief holds a picture's similarity data, as well as its equivalence set, and allows comparing pictures.
+ */
+class pic_equiv {
+public:
+	explicit pic_equiv(char const *cname);
+	~pic_equiv();
+	pic_equiv(const pic_equiv& other) = delete;
+	pic_equiv& operator=(const pic_equiv& other) = delete;
```
qarkai, line 42: I guess `pic_equiv` assignment operator could be restored using `shared_ptr` with custom deleter for `sim` member.

And one more thing. Would you mind renaming class to `PicEquiv` or something for naming consistency?

caclark:
@porridge 
Does the `--dupes-export` option allow you to do the actions you want?

PR #1526, opened 2024-09-30
Fix build on NetBSD.

0-wiz-0:
NetBSD does not provide statfs, only statvfs.

### DIFF
--- a/src/collect-io.cc
+++ b/src/collect-io.cc
@@ -51,6 +51,10 @@
 
 #define GQ_COLLECTION_MARKER "#" GQ_APPNAME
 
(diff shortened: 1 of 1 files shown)


caclark:
The compile FreeBSD check on GitHub loops with the message "VM is booting".

I have no idea where the problem is.

0-wiz-0:
The whole change is in an
```
#ifdef __NetBSD__
```
so FreeBSD shouldn't be affected.

caclark:
Implemented in commit f0ba2687bdad1a5dbc27c39330f47436e843a885

PR #1529, opened 2024-10-13
Check for GNU userspace instead of Linux kernel

gusnan:
`__linux__` checks only for Linux kernel, but this also is needed on other kernels - instead check if we are running on GNU userspace (`__GLIBC__`) which should work properly everywhere.

I noticed this because it failed to build on Hurd on the Debian autobuilders and this should fix this, and still build properly as previously on other ports.

### DIFF
--- a/src/collect-io.cc
+++ b/src/collect-io.cc
@@ -29,7 +29,7 @@
 #include <gdk/gdk.h>
 #include <glib-object.h>
 
-#if defined(__linux__)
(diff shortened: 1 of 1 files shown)


gusnan:
Some more info about the different defines here:  https://docs.gtk.org/glib/macros.html

caclark:
Implemented in commit 4709e028015630193d5a388651fe010dd42fc89f

Issue #1538, opened 2024-11-08
please consider use libraw instead ufraw 


sergiomb2:
**Setup (please complete the following information):**
 N/A 

**Describe your feature request**
ufraw is not updated since 2015 / 2018 and libraw have merged the ufraw code 

**Additional context**

I keep ufraw alive in https://github.com/sergiomb2/ufraw , but I think it is time to close , I saw in Fedora that geeqie  is the only thing depends on ufraw

sergiomb2:
https://github.com/LibRaw/LibRaw

qarkai:
libraw is already used, see https://github.com/BestImageViewer/geeqie/blob/47d3a069786a52a0dbe89e15e30ba2f0d03be854/meson.build#L447

IIUC ufraw was removed in https://github.com/BestImageViewer/geeqie/commit/04a1bff55ff08bd59ad07115f6abc2e3de31cca0

PR #1539, opened 2024-11-09
Remove ufraw strings from translations

qarkai:
Related to #1538


caclark:
Implemented in commit a2516722ceb84ea45fba97db138606808ab83859

PR #1547, opened 2024-11-30
Simplify FileData::marks_list_load a bit

qarkai:
Use g_auto macros.
Fix indentation.
Use for loop.
Reduce scope of variables and flatten code.

### DIFF
--- a/src/filedata/filedata.cc
+++ b/src/filedata/filedata.cc
@@ -2820,48 +2820,43 @@ static void marks_get_files(gpointer key, gpointer value, gpointer userdata)
 
 gboolean FileData::marks_list_load(const gchar *path)
 {
-	FILE *f;
-	gchar s_buf[1024];
-	gchar *pathl;
-	gchar *file_path;
-	gchar *marks_value;
+	g_autofree gchar *pathl = path_from_utf8(path);
 
-	pathl = path_from_utf8(path);
-	f = fopen(pathl, "r");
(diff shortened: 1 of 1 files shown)


src/filedata/filedata.cc
```diff
@@ -2820,48 +2820,43 @@ static void marks_get_files(gpointer key, gpointer value, gpointer userdata)
...
 		if (s_buf[0]=='#') continue;
-			file_path = strtok(s_buf, ",");
-			marks_value = strtok(nullptr, ",");
-			if (isfile(file_path))
+
+		const gchar *file_path = strtok(s_buf, ",");
+		if (!isfile(file_path)) continue;
+
+		const gchar *marks_str = strtok(nullptr, ",");
+		const gint marks_value = atoi(marks_str); // @fixme Skip if marks_value is 0?
```
qarkai, line 2845: @caclark @xsdg What do you think about this?

xsdg:
Agreed with @caclark 

Also, more broadly, I really like this change.  Thanks for continuing the good work 👍

src/filedata/filedata.cc
qarkai: Ok, added comment.

xsdg:
LGTM

caclark:
Implemented in commit 9c018368827c1e5650ac057c3d4480169584c801

PR #1549, opened 2024-12-01
Replace width/height enums with constants

qarkai:


caclark:
Is there any guidance on constant versus enum?

caclark:
Implemented in commit 8215cd64d2bb3ed35698e3a4ed704aa6f1bce9bb

PR #1550, opened 2024-12-03
Replace more width/height enums with constants

qarkai:


caclark:
Implemented in commit e6b84b32a914566cf362b9cddec101c507d0241c

PR #1551, opened 2024-12-04
Reduce number of const_cast

qarkai:


caclark:
Implemented in commit 86872b495c3129d0596bb950c3bbe1fc8ccaa81c

PR #1552, opened 2024-12-07
Fix g_strjoin memory leak

qarkai:
Also reduce duplication in translations.


caclark:
Implemented in commit dcf105b5c24d51d57de5850cc9508b7052757063

caclark:
@qarkai 
Thanks for the bug-cleaning...

PR #1553, opened 2024-12-07
Specify OsdTemplate type and simplify osd_template_insert()

qarkai:


caclark:
Implemented in commit db6271995d8d9e408329f47a50d198706ad3cf9d

PR #1554, opened 2024-12-09
Don't leak libheif image handles

ghakko:
When viewing an HEIF image without a color profile, `heif_color_profile()` in `color-man.cc` warns that reading the color profile failed, frees the libheif context and then returns, forgetting to free a libheif image handle that was just used. This leaks some memory and one file descriptor. When browsing many such images, the program will eventually crash after running out of file descriptors.

This fixes the leak and suppresses the warning in the very common case where the image has no color profile.

### DIFF
--- a/src/color-man.cc
+++ b/src/color-man.cc
@@ -699,6 +699,7 @@ guchar *heif_color_profile(FileData *fd, guint *profile_len)
 	if (error_code.code)
 		{
 		log_printf("warning: heif reader error: %s\n", error_code.message);
+		heif_image_handle_release(handle);
 		heif_context_free(ctx);
 		return nullptr;
 		}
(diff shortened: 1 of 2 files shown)


src/color-man.cc
```diff
@@ -699,6 +699,7 @@ guchar *heif_color_profile(FileData *fd, guint *profile_len)
 	if (error_code.code)
 		{
 		log_printf("warning: heif reader error: %s\n", error_code.message);
+		heif_image_handle_release(handle);
```
qarkai, line 702: Consider using [smart pointers](https://docs.gtk.org/glib/auto-cleanup.html) to avoid duplication of this block. Example: https://github.com/BestImageViewer/geeqie/blob/db6271995d8d9e408329f47a50d198706ad3cf9d/src/image-load-j2k.cc#L39C1-L41C62

qarkai: On second thought libheif has C++ API. It's probably a good time to use it.

caclark:
@qarkai  @ghakko 

I am sorry, but I lost track of this.

Is this OK as is?

qarkai:
@caclark It fixes resource leak, so OK. We can do polishing in other PRs.

caclark:
Implemented in commit 13b3f2e2938ceda0207806f750bfad037c7315bd

Sorry for the delay...

PR #1556, opened 2024-12-12
Add heif to mimetypes we can handle.

henning-schild:
If it can actually be handled will depend on config opts. But here we just align with i.e. image/tiff which will also be listed if not actually enabled.


caclark:
Implemented in commit 396faba834c70efa7c95a19dde52185c18959f5d

PR #1559, opened 2024-12-14
Use more g_autofree in ui-bookmark..windows

qarkai:
Fix several memory leaks.


caclark:
Implemented in commit 3d5ebe600769cadb86d60977c6b6d0fac22fdde2

PR #1560, opened 2024-12-15
Use pixbuf_scale_aspect() to reduce code duplication

qarkai:


caclark:
Implemented in commit 01b5153442aafcab2d553ffb7e6880e05379d666

PR #1563, opened 2024-12-19
Use g_list_find() to simplify code

qarkai:


caclark:
Implemented in commit 4e468b3c435e67f17ba6161b1f89defddf801d82

PR #1567, opened 2024-12-20
Replace GLib MIN/MAX macros with STL min/max functions

qarkai:
Use CLAMP where applicable.
Simplify related code.


caclark:
Implemented in commit 782f399c153f87eb8cf8d38ea986c8a6601ad87c

PR #1568, opened 2024-12-21
Use more g_autofree in pixbuf-util..trash

qarkai:
Drop some redundant string allocations.
Also flatten related code a bit.


caclark:
Implemented in commit 6bec48bef96418bc561b64743b6d3ca6b5555926

PR #1569, opened 2024-12-21
Use get_symbolic_link() to reduce code duplication

qarkai:
Flatten `pan_is_link_loop()`.
Also fix memory leak in `get_symbolic_link()`.


caclark:
Implemented in commit a0d1c068bf1090e4730d1ebbd7aa29cc0ca47fa6

PR #1570, opened 2024-12-22
Use more g_autofree in main..pan-view

qarkai:
Fix some memory leaks and simplify code.


caclark:
Implemented in commit 2ffa1fc2be575e683358876b1dadded6670a2f51

PR #1572, opened 2024-12-24
Replace custom gtkrc with gtk.css

qarkai:
Although not sure if anyone still uses it.


caclark:
Implemented in commit fa2fa073a4aa70e156cd08b5b5dd5c46dd8cd29a

PR #1576, opened 2024-12-26
Fix memory leak in layout_menu_kbd_map_cb()

qarkai:
Simplify a bit.


caclark:
Implemented in commit 0e82c56b856d46111abd95196cf44fe94a9bac9b

PR #1580, opened 2024-12-28
Use anonymous struct for keyboard_map_hardcoded

qarkai:
Also fix variables naming.


caclark:
Implemented in commit 222f27111a750d16a732b0f3ed88d60da2747c22

PR #1582, opened 2024-12-29
Inline trivial free-and-copy functions from image-overlay

qarkai:
Move `set_default_image_overlay_template_string()` to options.


caclark:
Implemented in commit 08232d09c2b749dd4a9a98637cb626bca6731229

PR #1583, opened 2024-12-29
Use more g_autofree in history-list..lua

qarkai:
Flatten and simplify code a bit.


caclark:
Implemented in commit 279c4a91f30b5eac6b0b9529a3964438929dc37b

PR #1585, opened 2024-12-30
Use more g_autofree in cache-loader..dupe

qarkai:
Fix some memleaks and simplify a bit.


caclark:
Implemented in commit 2804517bb7f16478c930ef71301480ad6519a8b4

PR #1586, opened 2024-12-30
Use more g_autofree in advanced-exif..bar-sort

qarkai:
Fix some memory leaks and simplify a bit.


caclark:
Implemented in commit d4babd3f258eb83c99c24cd020d605d09908d251

PR #1587, opened 2024-12-31
Simplify string array related code

qarkai:
Use more `g_auto(GStrv)` and `g_strv_*` functions.
Deduplicate uri-utils functions.


caclark:
Implemented in commit fe17beb2aa5d1a61da89d29d95f22dba1664139d

PR #1589, opened 2025-01-08
Simplify layout_menu_kbd_map_cb()

qarkai:
Split, deduplicate and flatten.


caclark:
Implemented in commit 9ad593831f77a5b2c026add23abbc27e1d9be3a8

PR #1590, opened 2025-01-09
Use g_autoptr(GError), fix several memory leaks

qarkai:
Deduplicate and flatten related functions.


caclark:
Implemented in commit 41ad5f5f7511f71753fed75eac766ac57868c95e

PR #1594, opened 2025-01-11
Use JxlDecoderPtr instead of hand-written unique_ptr

qarkai:
Also fix TODO.


caclark:
Implemented in commit db5572ca915043652d41c54d1ec867fc592c86b3

PR #1595, opened 2025-01-12
Use G_N_ELEMENTS instead of manual size calculations

qarkai:
Simplify related code.


caclark:
Implemented in commit c51ae47d8a8ccb78023d7acc7a672184095131e3

PR #1596, opened 2025-01-13
Fix memory leak after exif_get_data_as_text() call

qarkai:
Reduce code duplication in related functions.
Also fix '_XOPEN_SOURCE redefined' warnings.


caclark:
Implemented in commit f428cc618e4a796aa9adbf67d504a7eab544c98c

PR #1597, opened 2025-01-14
Simplify jpeg-parser a bit

qarkai:
Comment out unused members in `MPOEntry` and clarify FIXME.
Replace pointers with references.
Use explicit types for parameters in parse callbacks.
Flatten `jpeg_get_mpo_data()`.


caclark:
Implemented in commit dbad0be2344767b8784d28ea59903c56aff5a692

PR #1601, opened 2025-01-16
Add TiffTag struct to reduce code duplication

qarkai:
Also move static functions to anonymous namespace.


caclark:
Implemented in commit 20deb477ed4f56e50abc2df8f445a6ab09478d7b

PR #1602, opened 2025-01-16
Remove redundant members from ExifWin struct

qarkai:


caclark:
Implemented in commit 4161e76ded577aa81dcc5215422ded2dfbf1d042

PR #1603, opened 2025-01-17
Avoid manual memory management in jpeg-parser

qarkai:
Use vector for `MPOData::images`.
Change return type of `jpeg_get_mpo_data()`.

### DIFF
--- a/src/jpeg-parser.cc
+++ b/src/jpeg-parser.cc
@@ -21,6 +21,7 @@
 
 #include "jpeg-parser.h"
 
+#include <algorithm>
 #include <cstring>
 #include <functional>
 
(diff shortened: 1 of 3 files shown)


src/jpeg-parser.cc
```diff
@@ -161,30 +162,23 @@ gint mpo_parse_Index_IFD_entry(const guchar *tiff, guint offset, guint size, Tif
...
+			MPOEntry &mpe = mpo.images[i];
 
-			DEBUG_1("   image %x %x %x", image_attr, mpo.images[i].length, mpo.images[i].offset);
+			guint image_attr = tiff_byte_get_int32(tiff + data_offset + (i * 16), bo);
+			mpe.type_code = image_attr & 0xffffff;
+			mpe.representative = !!(image_attr & 0x20000000);
+			mpe.dependent_child = !!(image_attr & 0x40000000);
+			mpe.dependent_parent = !!(image_attr & 0x80000000);
+			mpe.length = tiff_byte_get_int32(tiff + data_offset + (i * 16) + 4, bo);
+			if (i > 0) mpe.offset = tiff_byte_get_int32(tiff + data_offset + (i * 16) + 8, bo) + mpo.mpo_offset;
```
xsdg, line 177: Might be good to add a single-line comment that mpe.offset is zero-initialized, so that it's clearer that this isn't creating uninitialized memory when `i==0`

qarkai: Is it enough to add explicit default values to `MPOEntry` members instead?

xsdg: I was thinking that the purpose of the comment is to avoid having to even look at how MPOEntry is defined.  From what I looked up, the standard requires the behavior that this code relies on, so that doesn't feel like as much of a concern (though feel free to add more explicit initialization if you'd like).  I was just thinking that having a comment here would just help avoid confusion for someone reading this code.

FTR, I was looking at these refs for standards behavior:
 - https://stackoverflow.com/questions/48775464/why-does-c-initialise-a-stdvector-with-zeros-but-not-a-stdarray
 - https://stackoverflow.com/questions/4932781/why-is-a-pod-in-a-struct-zero-initialized-by-an-implicit-constructor-when-creati

qarkai: Changed a bit. What do you think?

qarkai: FYI Here is the diff https://github.com/BestImageViewer/geeqie/compare/f74c4d7166c5be7d1bed4229146cc27b2b2cbda1..9f934c94783ea77d686dbfdfd2a258c7c52d7123

caclark:
Implemented in commit ccb88a002b226a85aa41f2d9003fc4e357b597f5

PR #1604, opened 2025-01-18
Fix #1591: Allow drawing rectangle upwards and to the left

### LINKED ISSUE #1591: Can't draw rectangle from a lower corner
Tried drawing a rectangle upwards and was met with a line instead of a rectangle.

Before I start working on this bug, the relevant code is in `image_drag_cb()` in `image.cc` ? Is there any reason why (in my simple non-C++ programmer terms) I just couldn't swap the values of [`pixbuf_start_x/y values` here](https://github.com/BestImageViewer/geeqie/blob/bc5e15ef239750a9d1265787cd5690adef8c02af/src/image.cc#L294) and `pr->drag_last_x` when the difference between the drag and the pixbuf_start_x/y is going negative? I might change the naming to `top_left/right` and `bottom_left/right` to make it a little clearer; and use `CLAMP()` to stay within the image and aspect ratio constraints.

Any thoughts before I dig into it over the weekend?

TomK32:
Finally got around to sort my testing out, no further changes to the rectangle computation I mentioned in the issue ticket were required :-)

### DIFF
--- a/src/image.cc
+++ b/src/image.cc
@@ -55,12 +55,38 @@ static void image_read_ahead_start(ImageWindow *imd);
 static void image_cache_set(ImageWindow *imd, FileData *fd);
 
 // For draw rectangle function
-static gint pixbuf_start_x;
-static gint pixbuf_start_y;
 static gint image_start_x;
 static gint image_start_y;
 static gint rect_x1, rect_x2, rect_y1, rect_y2;
 static gint rect_id = 0;
+static SelectionRectangle selection_rectangle;
+
+/*
+ * SelectionRectangle
+ */
+
+void SelectionRectangle::setCursor(gint x, gint y) {
+	this->cursor_x = x;
+	this->cursor_y = y;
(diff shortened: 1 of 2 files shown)


caclark:
In `image.h` your editor has sometimes used spaces instead of tabs. Would you please use tabs.

When running from the command line, I see many warnings such as
`(geeqie:32171): GdkPixbuf-CRITICAL **: 10:50:45.803: gdk_pixbuf_calculate_rowstride: assertion 'height > 0' failed`

I see that at line 279 of `image.cc` (in the modified file) width or height are sometimes zero.

TomK32:
Oh yes sorry. tabs is something we Ruby developer don't do, but that `expandtab` setting in my vimrc probably never was a good idea anyways. I fixed the indentation and also set height/width to 1 if they are too small.

caclark:
Implemented in commit 45b55848b909e8e9d34829bf286aea654a2eb03a

PR #1609, opened 2025-01-25
Simplify ImageLoaderJpeg::write() a bit

qarkai:
Avoid some casting.
Use iterators instead of indexes.
Filter images before checking size.


caclark:
Implemented in commit 19c69f65d7ba2c36d95bdeb971cfe981d42bbf7a

PR #1611, opened 2025-01-25
Use libheif C++ API

qarkai:
Simplify `heif_color_profile()` and related a bit.


caclark:
Implemented in commit 590a5c2fb02f8933b3c9f263b30a089b5c2e6324

PR #1612, opened 2025-01-26
Split heif_color_profile() to separate module

qarkai:
Simplify conditional compilation.


caclark:
Implemented in commit 3e63f92826b172b784044bb0fac9d362dd5ed8c8

PR #1613, opened 2025-01-28
Move lcms dependent code from exif-common to color-man

qarkai:


caclark:
Implemented in commit bc481e78f3a3c53da7c7541200fb7514c345f4f8

PR #1616, opened 2025-01-30
Reduce number of warnings

qarkai:
So we do not miss newly added warnings.

Silence all deprecation warnings.
Move aliases for deprecated symbols to separate `compat-deprecated.h` file
to avoid false positives in GTK4 migration test.

Fix existing warnings.


xsdg:
Just one opinion: \
I'm not generally a fan of using `-Werror` as a default option, because it means that whether or not the codebase compiles ends up being really sensitive to which specific compiler version someone is using.  So, for example, if Debian ships a new compiler in unstable that has a new warning, then geeqie will stop building, which means that Debian users on testing or stable will stop getting new versions until we (or the Debian maintainers) fix or disable that new warning.

I do think it's good to turn it on from time to time just to check what we might be missing, but my personal opinion is that it should be off by default.

qarkai:
> So, for example, if Debian ships a new compiler in unstable that has a new warning, then geeqie will stop building, which means that Debian users on testing or stable will stop getting new versions until we (or the Debian maintainers) fix or disable that new warning.

Good point. I'll revert `-Werror`.

caclark:
How about -DWerror=enabled option?

qarkai:
IIUC `werror=true` and `-DWerror=enabled` is the same.

PR #1617, opened 2025-02-02
Add exif_get_color_profile() to reduce code duplication

qarkai:


caclark:
Implemented in commit 4cae35dee5340b93c36d8554fae0470ffd882acf

qarkai:
@caclark @xsdg I noticed that logic related to `Exif.Iop.InteroperabilityIndex` is slightly different in [`exif_build_formatted_ColorProfile()`](https://github.com/BestImageViewer/geeqie/blob/4cae35dee5340b93c36d8554fae0470ffd882acf/src/exif-common.cc#L422) and newly added [`exif_get_color_profile()`](https://github.com/BestImageViewer/geeqie/blob/4cae35dee5340b93c36d8554fae0470ffd882acf/src/exif-common.cc#L937).
`exif_build_formatted_ColorProfile()` doesn't call `name = _("sRGB");` when `interop_index` is `"R98"`.

Is it intentional or these two have to be similar?

caclark:
This does not look correct to me - but I do not understand the exif stuff.

My opinion is that the sequence looking for color data should be  (take the first hit and ignore conflicts):
1. Exif.Image.InterColorProfile (embedded color profile)
2. Exif.Photo.ColorSpace
3. Exif.Iop.InteroperabilityIndex
... and the get_profile() should match the get_formatted_profile() code.

(That is just an opinion - I did not find a rule for that)

PR #1618, opened 2025-02-04
Drop unused layout_new()

qarkai:
Rename `layout_new_with_geometry()` to `layout_new()` and make static.
Simplify new `layout_new()` parameters.


caclark:
Implemented in commit f7fcd348a2cffac02bbdfe9d5fed9c0510bd6d90

PR #1621, opened 2025-02-06
Improved some translations to Italian language

pezcurrel:
As the title says :-) (I guess i won't touch it again until some changes in the sources will require it)


caclark:
Implemented in commit c1785ca550bf4bd450d8a55d7ce670d174ad445c

PR #1622, opened 2025-02-06
Use menu_item_* functions to reduce duplication

qarkai:


caclark:
Implemented in commit b85c8d39169c426ea62dba8958410967388c172d

PR #1623, opened 2025-02-06
Use vector for view_window_list

qarkai:
Also harden `view_window_find_image()` parameters a bit.


caclark:
Implemented in commit 113a36eda86c011b6124eb9f82016701598acddd

PR #1624, opened 2025-02-07
Simplify class_filter_menu() callbacks

qarkai:
We already know state by menu item, no need to calculate it each time.


caclark:
Implemented in commit 21502a2828b789389e683d4dcc0fda232b219b66

PR #1625, opened 2025-02-08
Hide layout_window_list

qarkai:


caclark:
Implemented in commit 6cf6431573c10daa03260ce0b76bd3f14c0ec759

PR #1626, opened 2025-02-08
Use gresources for keymap template

qarkai:


caclark:
Implemented in commit fb05084a5ca40cae107b4eb856a027c2b21ab981

PR #1628, opened 2025-02-09
Use vector for layout_window_list

qarkai:


caclark:
Implemented in commit fa8bce2ab1c799da405f7111729a05fb22bf51e5

PR #1629, opened 2025-02-09
Use gresources for icc file

qarkai:
Drop xxd build requirement.


caclark:
Implemented in commit a395933b8847ba77c943e210b5a3a0865422953e

PR #1630, opened 2025-02-10
Drop redundant HAVE_LCMS2 flag

qarkai:
lcms required version is >=2.0.
Remove obsolete `color_man_lib_init()`.

### DIFF
--- a/config.h.in
+++ b/config.h.in
@@ -86,9 +86,6 @@
 /* color profiles with lcms */
 #mesondefine HAVE_LCMS
 
-/* Define to enable use of lcms2 */
-#mesondefine HAVE_LCMS2
-
 /* Define to enable use of maps */
 #mesondefine HAVE_LIBCHAMPLAIN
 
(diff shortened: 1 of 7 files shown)


qarkai:
@caclark I don't know why tests failed. I have tried to open failed images locally, no issues. Could you help me here?

caclark:
I get occasional test fails on GitHub - especially when I am not at home and logging in via some other method.
My first step is to rerun the test - it usually passes, and I never bothered to analyse why the first run failed.

Sometimes it fails the second time as well - that is because I have made a mistake....

I have rerun the test and it is OK.

caclark:
@qarkai 

The test fails were on image tests. I think there may be a timing sensitivity somewhere in the test scripts.
I will check that sometime.
Meanwhile rerunning the test usually works.

caclark:
Implemented in commit e77c15003e41dd6a6e4dbd1f765267207aac7d88

PR #1631, opened 2025-02-11
Simplify and inline find_string_in_list_utf8* functions

qarkai:
Change `find_string_in_list()` return type since
`g_list_delete_link()` is faster than `g_list_remove()`.


caclark:
Implemented in commit 302c420c32af357419e238588ac760206bdd1ce0

PR #1633, opened 2025-02-12
Simplify collection_load_private() a bit

qarkai:
Split check if file was on a mounted drive.
Remove redundant `reading_extended_filename` flag, use `extended_filename_buffer` instead.
Fix memory leak in case of invalid collection file.
Flatten code.


caclark:
Implemented in commit 418b252cb0008162b619491d2047356016da780e

PR #1634, opened 2025-02-13
Use g_autoptr in is_file_on_mounted_drive()

qarkai:
Deduplicate 'is file in directory' logic and reorganize code a bit.


caclark:
Implemented in commit efa349945fefdbdd52bcf7c7cfce2584459ab774

PR #1636, opened 2025-02-16
Simplify help_search_window signals handling

qarkai:
Remove redundant `HelpSearchData` struct.


caclark:
Implemented in commit ddc5191c52fe1a378810c7e9137491b3aa3e49bb

PR #1638, opened 2025-02-16
Remove unused GtkWidget parameter from defined_mouse_buttons()

qarkai:
Also deduplicate code.


caclark:
Implemented in commit f410bd7

PR #1639, opened 2025-02-17
Use text_widget_text_pull() to reduce duplication

qarkai:
Also fix memory leaks.


caclark:
Implemented in commit 9bf21227bcf94600c2bbe28bc6f1782a38a76387

PR #1640, opened 2025-02-18
Fix memory leak after gtk_text_buffer_get_text() call

qarkai:
Simplify `search_keypress_event()` a bit.
Also use `text_widget_text_pull()` in one more place.


caclark:
Implemented in commit d863ff031731d41055b90d545e16869bd9027d96

PR #1641, opened 2025-02-18
Fix memory leak after keyword_list_get() call

qarkai:
Simplify `config_tab_keywords()` a bit.


caclark:
Implemented in commit 0b63ea9f889feeb669aac920d9333c67e19dd971

PR #1646, opened 2025-02-22
Add G_GNUC_PRINTF to secure_fprintf()

qarkai:
Simplify `secure_fprintf()` calls and fix memory leak.


caclark:
Implemented in commit 5efb02b01406e5b092d9a56df51f46d034e5f628

PR #1649, opened 2025-02-23
Simplify copy of options

qarkai:


caclark:
Implemented in commit e0657b35908cc90f2751de73f7c526630cd96ded

PR #1651, opened 2025-02-26
Use Meson GNOME module

qarkai:
Split code generation from `authors.sh` and `translators.sh` scripts.


caclark:
Implemented in commit 2d47dac189f6423220aeb4bca7f816cdbbcfa2fa

PR #1652, opened 2025-02-27
Remove unused and read-only members of CommandLine struct

qarkai:
Rename `ssi` to `log_file_ssi`.
Simplify related code and fix memory leak.


caclark:
Implemented in commit 9ccaa7db2322176ccbcd93d505cd7e77d14a16aa

PR #1653, opened 2025-02-27
Fix set_regexp() memory leaks

qarkai:
Also move `gq_debug()` and `gq_grep()` under `DEBUG` macro.


caclark:
Implemented in commit eb39809c0ddd7c3fe5f9c0864b1398954905f9d9

PR #1658, opened 2025-03-01
Do not generate existing marshallers

qarkai:


caclark:
Implemented in commit d0857fe632df329c64a3f4f14a910b7c853c3c5f

PR #1661, opened 2025-03-01
Simplify image_select() call a bit

qarkai:
Use `bool` for `select` parameter.


caclark:
Implemented in commit 639db881488f8a7b886562c34af6ae96575ba23c

PR #1662, opened 2025-03-01
Avoid gdouble compare in SelectionRectangle

qarkai:


caclark:
Implemented in commit 1e3059b4dc20a43899a54c4d62515d77a2291b5e

PR #1663, opened 2025-03-02
Move scale_factor() to color-man

qarkai:
The only place where it is called.


caclark:
Implemented in commit fe9bd87d530d511834099a2ba9e327e05cd82fa9

PR #1666, opened 2025-03-03
Fix memory leak

qarkai:


caclark:
Implemented in commit 2aee1ba301d0db0b70a6bda3e5c13b9bf99fde5d

PR #1667, opened 2025-03-03
Move NotifyType enum to filedata

qarkai:


caclark:
Implemented in commit f638b0196f007fb765daad9a15d00c30d63f922a

PR #1668, opened 2025-03-04
Simplify histogram

qarkai:
Move `Histogram` grid to struct, use constant instead of members.
Use `guint8` for grid color components.
Remove return types from `histogram_set` functions.
Make some parameters `const`.


caclark:
Implemented in commit 97bb69f1fd94b20eecf548fe0a6a99597b7cc2d8

PR #1669, opened 2025-03-04
Move logwindow variables declarations and callbacks under DEBUG macro

qarkai:


caclark:
Implemented in commit d9ffcdb2cf105d9e36f89ed56ffecf30bb68f284

PR #1670, opened 2025-03-05
Simplify nclx_to_lcms_profile()

qarkai:
Use g_auto* macros.
Reduce code duplication.
Flatten code.


caclark:
Implemented in commit 8e04ce91565f4f46f80a654308a127783a7d8a31

PR #1671, opened 2025-03-06
Add explicit values for Histogram::histogram_mode

qarkai:
Also convert `HCHAN_*` macros to enum.


caclark:
Implemented in commit 3177ca4b86cffb4661b0a4f4c474b7fc8f894a8f

PR #1672, opened 2025-03-08
Convert histogram members from pointer to value

qarkai:
Add methods to `Histogram` class to simplify code.
Change `Histogram::draw()` return type.


caclark:
Implemented in commit bd6d298da0918f59210a18854fbdebb989035b26

PR #1673, opened 2025-03-08
Replace some unlink() calls

qarkai:
Use `std::remove()` or `unlink_file()`.
This allows to reduce dependency on `unistd.h`.


caclark:
Implemented in commit 340c6fa9f48775b659874c95811cf031701c4718

PR #1674, opened 2025-03-08
Fix build without debug

qarkai:


qarkai:
@caclark What do you think about adding no-debug check builds, i.e. with `-Ddebug=false`? E.g. add `Check-Build-No-Options-Release` and `Check-Build-Most-Options-Release`.

caclark:
Somewhere in the docs it says that Geeqie is "stable" - but that just means that it does not fall over every day. There will always be bugs.

The disadvantage of a no-debug release is that if a user finds a problem that is reproducible, it would not be possible to ask the user to re-run with `--debug=4`.

Would a no-debug version run faster?

If speed is the problem, maybe `perf` could be used to identify where. (At the moment my son is working on `perf` for Arm - but I have no understanding of what he does...)

qarkai:
I don't mean distributing release versions, just check if no-debug build is successful and doesn't have warnings maybe.

AFAIU some downstream packages use release builds. E.g. Arch Linux uses [`--buildtype plain`](https://mesonbuild.com/Builtin-options.html#details-for-buildtype) for all Meson packages.

caclark:
Implemented in commit 6dca0ec605eb7e98691a0f506b58aaf4b877f002

caclark:
OK, I understand - maybe.

If it is just a matter of checking compiles with and without the debug option, would it be enough to put `-Ddebug=false` in the `Check-Build-No-Options:` section of `check-build-actions.yml` file?

[`Arch Linux uses --buildtype plain` Creating a release with debug = no optimisation seems strange]

qarkai:
> would it be enough to put `-Ddebug=false` in the `Check-Build-No-Options:` section of `check-build-actions.yml` file?

No, need to check build without debug for Most-Options too since Most-Options build different files than No-Options. Let's continue in #1675.

PR #1677, opened 2025-03-12
Add HistoryChain struct to reduce code duplication

qarkai:
Use vector for chain data.


src/history-list.cc
```diff
@@ -31,6 +32,67 @@
 namespace
 {
 
+struct HistoryChain
+{
+	const gchar *back();
```
xsdg, line 37: It'd be good to mark these accessor methods as `const` so that they can still be called on a `const HistoryChain` if needed at some point.

So `const gchar *back() const;`, I believe.

xsdg: Oh, disregard 🙃

xsdg, line 37: In hindsight, part of the confusion around my last comment is that this API feels like it's not aligned with the STL APIs.  In particular, there is a pattern of `back` being a non-modifying method that returns a reference to an element at that location in the container.  For example: \
https://en.cppreference.com/w/cpp/container/vector/back

Given that these are methods that are intended to change the state of the `HistoryChain` container, it would be good to use method names that don't conflict with how those same names are used within the STL.

As two ideas (just ideas; probably not the best), maybe `rewind` instead of `back`, and `advance` instead of `forward`?

Likewise, it feels like `append_end` should really just be `push_back`

qarkai: Probably `next` and `prev` are what you were looking for.

xsdg: lgtm 👍

src/history-list.cc
```diff
@@ -31,6 +32,67 @@
...
+	std::vector<gchar *> chain;
+	guint index = 0;
+	bool is_nav_button = false; /** Used to prevent the nav buttons making entries to the chain **/
+};
+
+const gchar *HistoryChain::prev()
+{
+	is_nav_button = true;
+
+	index = index > 0 ? index - 1 : 0;
```
xsdg, line 51: Completely up to you, but `index = std::max(0, index - 1);` might be easier to follow (and likewise on line 60/61)

qarkai: If `index` is 0, then `index - 1` is `G_MAXUINT`.

xsdg: Aah, yes, I forgot that it's unsigned.  Thanks for the correction 🙏

caclark:
Implemented in commit 72cb54ee50979feeeb170c5d0a9058c335c69566

PR #1678, opened 2025-03-13
Fix memory leaks after gtk_tree_selection_get_selected_rows()

qarkai:


caclark:
Implemented in commit b19dfb05e1ba99825d4c9233f58c454140e9ab51

PR #1679, opened 2025-03-14
Use gtk_tree_selection_count_selected_rows()

qarkai:
Reduce scope of variables in `vflist_selection_count()`.


caclark:
Implemented in commit 7ecb36ed43d38f1f0f75489ddde10f06be56a66b

PR #1681, opened 2025-03-16
Make help_window local static variable

qarkai:


caclark:
Implemented in commit dcfe4e8049a3407fc126d2a5d8e3bf9b739c6d79

PR #1682, opened 2025-03-16
Simplify selection_done signal handling

qarkai:


caclark:
Implemented in commit 65c7468368dcf46d50e5f0bb2eca0cccdcfde00b

PR #1686, opened 2025-03-20
Drop redundant SelectionRectangle::cursor_x/y

qarkai:


caclark:
@qarkai 
I cancelled the FreeBSD run because was clear there was a problem with GitHub's VM, and it was stalled.

I now cannot get it to rerun. Sorry.

caclark:
Implemented in commit f0ba81bfb1639bd5d21627e93d0bc1ecf897459e

PR #1687, opened 2025-03-22
Hide editors variable

qarkai:


caclark:
Implemented in commit 24decbec0a807b91d0b2a8c183c2316ff513d77e

PR #1688, opened 2025-03-22
Use vector for editors list

qarkai:
Simplify related code.


caclark:
Implemented in commit be747802d06886f256ac60a8fc5fbe64700f0f43

PR #1689, opened 2025-03-22
Drop unused indent parameter from write_*_option() functions

qarkai:
Change `write_bool_option()` parameter to `gboolean`.
Add `const` to `write_color_option()` parameter.


caclark:
Implemented in commit 17378ca6b5c2deda3b20f1ab45c45c6b275ded30

PR #1690, opened 2025-03-23
Fix memory leak after gtk_widget_get_tooltip_text() call in GTK3

qarkai:
Also get `button_list` once.

### DIFF
--- a/src/layout-util.cc
+++ b/src/layout-util.cc
@@ -3072,53 +3072,44 @@ static void layout_actions_setup_editors(LayoutWindow *lw)
 		g_string_append(desc, "    <menu action='OpenMenu'>");
 		}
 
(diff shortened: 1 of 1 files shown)


caclark:
Implemented in commit 1d73d8dae5ef756699e2a8e8526ad8369fe8dab7

caclark:
@qarkai 

Running `G_DEBUG=fatal-warnings ./build/src/geeqie` on this commit
gives
`(geeqie:7244): Gtk-CRITICAL **: 14:18:26.890: gtk_container_foreach: assertion 'GTK_IS_CONTAINER (container)' failed
Segmentation fault (core dumped)`

Sorry - it is not clear to me where the problem is.

qarkai:
@caclark Quick fix is in #1693. Probably `layout_actions_setup_editors()` should not be called at all before main toolbar is initialized.

PR #1694, opened 2025-03-24
Inline shortcuts_add_close()

qarkai:
Also remove unused includes.


caclark:
Implemented in commit b38f92f626e4e60cae21f28ccd11342b93f09930

PR #1695, opened 2025-03-24
Drop unused parameter from save_default_layout_options_to_file()

qarkai:


caclark:
Implemented in commit d1e51783d337b779cb2b00aa780bd5dc814c66d8

PR #1696, opened 2025-03-26
Enable clang-tidy misc-header-include-cycle check

qarkai:
Add third-party headers to ignore list.


caclark:
Implemented in commit f1c0fde9394868268a41ea8e7169fe81898e15c6

PR #1700, opened 2025-04-02
Simplify code related to ViewFile stars

qarkai:
Move common code to `vf_star_next()`.
Make `vf_stars_cb()` local static and flatten a bit.
Drop redundant `vf_star_do()`.


caclark:
Implemented in commit 297dcdb1babb105549e76956d29d803f1d851108

PR #1702, opened 2025-04-05
Cleanup GSourceFunc implementations

qarkai:
Remove redundant casts.
Use `G_SOURCE_CONTINUE`/`G_SOURCE_REMOVE` instead of `TRUE`/`FALSE`.
Make `metadata_write_idle_id` local static variable.
Reduce code duplication.


caclark:
Implemented in commit 69da0e66183a5be61f938fa6657c7848b4cf9d95

PR #1704, opened 2025-04-07
Simplify check for parent window

qarkai:


caclark:
Implemented in commit d4f9a0445fead105cdded4c4b6a750b0cfe13b4e

PR #1705, opened 2025-04-09
Use enum class for UtilityPhase and UtilityType

qarkai:


caclark:
Implemented in commit 604c965828b251fa046b27986ea361c12dc09570

PR #1710, opened 2025-04-12
Fix memory leaks after vf_file_filter_get_filter() calls

qarkai:
Use `g_autoptr(GRegex)`.


caclark:
Implemented in commit 21b296b196fb8aceb5cd657af51b88c9cb530d6d

PR #1711, opened 2025-04-12
Make GRegex parameters const

qarkai:


caclark:
Implemented in commit 2708a16b6326e96c2184d79428f9fe8931493c09

PR #1713, opened 2025-04-12
Reduce code duplication in pan_filter_toggle_cb()

qarkai:


caclark:
Implemented in commit 8014c6db27f99b1ebd2d69eb01f2f6220b863084

PR #1715, opened 2025-04-12
Move pan-view-filter local types and functions to anonymous namespace

qarkai:


caclark:
Implemented in commit bdf6c78d9a3d6a3df59bdebd63dffbe0352d4c01

PR #1716, opened 2025-04-12
Fix memory leak of PanViewFilterElement

qarkai:


caclark:
Implemented in commit 17e87530a7360376254515eab28554b8959cdb31

PR #1719, opened 2025-04-14
Fix more memory leaks in pan-view-filter

qarkai:


caclark:
Implemented in commit dbe72a89a24703bd402de33a94918f8fc86c1900

PR #1724, opened 2025-04-19
Drop unused parameters from print_window_new()

qarkai:


caclark:
Implemented in commit d807467daaf99feb39301c0660f429bc06429b83

PR #1725, opened 2025-04-20
Replace mfd_list_free() with g_list_free_full()

qarkai:


caclark:
Implemented in commit e10708d0f7d512a07a02677cdeff6d21a4fecdb7

PR #1726, opened 2025-04-20
Replace defines in GPS match with constexpr

qarkai:
Simplify a bit.


caclark:
Implemented in commit d17d139b583047f968cdc5f1e961508609306a8d

PR #1727, opened 2025-04-22
Fix gtk_combo_box_text_get_active_text() memory leaks

qarkai:


caclark:
Implemented in commit fbd19d8a87467d112490a5951c4b1b39706e972b

PR #1728, opened 2025-04-23
Replace MATCH_IS_BETWEEN macro with template function

qarkai:


caclark:
Implemented in commit 5ce31ddf9b11b852176a75f1ebc9c01e22172d5e

PR #1729, opened 2025-04-25
Drop redundant SearchData::button_help member

qarkai:


caclark:
Implemented in commit cdc768eaee066e870c10521781b82ce2ac6def5b

PR #1730, opened 2025-04-26
Fix g_strdup_printf() memory leaks in search

qarkai:


caclark:
Implemented in commit 822d08e247558eece3593bccb541e7341e93b7e7

PR #1732, opened 2025-04-26
Split search UI from data

qarkai:


caclark:
Implemented in commit c5ec455560062f53ab97bc30a2a96228c239984e

PR #1734, opened 2025-04-27
Add SearchDate to reduce code duplication

qarkai:
Set values only if match date enabled.


caclark:
Implemented in commit 6a3ddc9408d54c9307c1f4e67b09460663adf554

PR #1735, opened 2025-04-27
Simplify calculation of search date boundaries

qarkai:


caclark:
Implemented in commit 9a7a6413e4d1bc426f367ede812f190e9e5b4194

caclark:
@qarkai 
When running tests locally, I now get this error:
`../src/search.cc:2075:47: error: '*' has higher precedence than '+'; add parentheses to explicitly specify the order of operations [readability-math-missing-parentheses,-warnings-as-errors]
 2075 |                         match = match_is_between(file_date, a, b + 60 * 60 * 24 - 1);`

Has something changed somewhere, or is your setup different to mine?

qarkai:
@caclark I get this error too but I don't run tests regularly.
Main question is why it wasn't detected in GitHub Checks.
Also why `match = (file_date > sd->search_date.to_time() + 60 * 60 * 24 - 1);` doesn't trigger this warning.

caclark:
So far as I can tell:

It was included in 19.1.0
`https://releases.llvm.org/19.1.0/tools/clang/tools/extra/docs/ReleaseNotes.html`

Github does not have that version:
`https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md`

PR #1736, opened 2025-04-28
Fix date type search

qarkai:
And determine required `FileData` datetime member once in `search_start_cb()`.


caclark:
Implemented in commit f41d8d8db40eaca85f10b03b6c9299ce95beec40

PR #1737, opened 2025-04-29
Drop unused search_window_list

qarkai:


caclark:
Implemented in commit 8a611878596f7f6dc01ba54badbd1860a4c3d4a7

PR #1738, opened 2025-04-29
Get search class once per search

qarkai:


caclark:
Implemented in commit 3334f7028b1bb17165e51180664989aeec4e53c1

PR #1739, opened 2025-04-30
Fix build with GCC 15

qarkai:


caclark:
Implemented in commit ea7a2e30a4f28688b511cc188ebdb6622ed5889c

PR #1740, opened 2025-04-30
Get search marks once per search

qarkai:


caclark:
Implemented in commit 7eaa9a58c7284b7b2cccd818c09ea27d71274bd9

PR #1741, opened 2025-05-05
Fix readability-math-missing-parentheses warning

qarkai:
Found in https://github.com/BestImageViewer/geeqie/pull/1735#issuecomment-2850386733.


caclark:
Implemented in commit 7d1037809b672b0ad6cd4c8647f5d8ed67e7c908

PR #1742, opened 2025-05-06
Get search gps once per search

qarkai:
Fix search gps if "Image is not geocoded" is selected.


caclark:
Implemented in commit defb703b05f00590796f00f103f5935ba357e4a2

PR #1744, opened 2025-05-07
Move common code to get_marks_string()

qarkai:


caclark:
Implemented in commit 1d7a8785f664c8ce3c61fd6325b82e59ed03a98d

PR #1751, opened 2025-05-24
Drop unused lw parameter from layout_selectable_toolbars_toggle()

qarkai:
Rename to `current_layout_selectable_toolbars_toggle()`.


caclark:
Implemented in commit cd0309d838235e3b42e0ab95a0f8ea23f4ce4029

PR #1753, opened 2025-05-25
Add FileDataList alias for GList with FileData elements

qarkai:
Rename `filelist_free()` to `file_data_list_free()`.
Use `g_autoptr` macro to simplify code a bit.


caclark:
Implemented in commit e6b779b04fbd9b39a73a8c8d45d3b0d5216bdee1

PR #1757, opened 2025-05-27
Move secure_save() to ui-fileops

qarkai:
Fix group id getting.
Simplify a bit.


caclark:
Implemented in commit 7348cc066dfa9c02094fe812b25f6a4adf18446d

PR #1769, opened 2025-06-07
Use gtk_container_foreach() to simplify code

qarkai:
Fix couple memory leaks.


caclark:
Implemented in commit 90033d5342a7826

PR #1775, opened 2025-06-12
Add string_list_join() to reduce duplication

qarkai:


caclark:
Implemented in commit 304d60583ad5de61462cb80e76b96a10519dbcca

PR #1776, opened 2025-06-12
Simplify TagData handling

qarkai:
Remove unused `title` member.
Add `image_overlay_template_view` member to avoid using GObject properties.
Use `g_signal_connect_swapped()` to simplify callbacks.
Move callbacks to anonymous namespace.


caclark:
Implemented in commit 82def3afdf5270ee54abdfef723cd2a326debc98

PR #1781, opened 2025-06-19
Stable/2.6

MPolleke:
Update Dutch (nl) translation.
Fix 3 typing mistakes. Update po files too, though I didn't dare to touch ko.po.


caclark:
@MPolleke 
Thanks for the patch.

Unfortunately the patch fails - I think `./src/shortcuts.cc` has changed since you created the patch.

caclark:
Something is amiss...

There is commit ea27e5d 2025-05-05 which changes `./src/shortcuts.cc`.

Your commit is in the correct place - geeqie.org and the github geeqie repos should be in sync.

You should make your patches against HEAD, rather than against a version tag.
Just `git clone <etc>` and work from there.

PR #1784, opened 2025-06-21
Fix pdf formats typo

qarkai:
Looks like `formast` fixed in 3491a34870c4f0ce89c9e515e0f0decf0693e6bf was a typo for `format`, not `formats`.


caclark:
Implemented in commit 2fb84d63837fa418ed2f421f12e1022c7989e767

PR #1790, opened 2025-06-26
Fix memory leaks after layout_window_menu_list() calls

qarkai:
Also remove redundant parameter.

### DIFF
--- a/src/layout-util.cc
+++ b/src/layout-util.cc
@@ -2266,6 +2266,15 @@ struct DeleteWindow
 	GtkWidget *group;
 };
 
+static void window_names_free(WindowNames *wn)
+{
+	if (!wn) return;
+
+	g_free(wn->name);
+	g_free(wn->path);
+	g_free(wn);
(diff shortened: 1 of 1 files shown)


qarkai:
@caclark Is it possible to bump GLib required version to at least 2.56? It introduced `g_autolist` macro which allows to write
```cpp
G_DEFINE_AUTOPTR_CLEANUP_FUNC(WindowNames, window_names_free)
.
.
.
g_autolist(WindowNames) list = layout_window_menu_list();
```
instead of
```cpp
GList *list = layout_window_menu_list();
.
.
.
g_list_free_full(list, reinterpret_cast<GDestroyNotify>(window_names_free));
```

This approach has two advantages:
1. Avoid manual memory management
2. View `GList`'s data type. Although still no compile-time type checks.

caclark:
Implemented in commit 24643b8503d84f4d6d455687b7ed2293cdab139f

PR #1791, opened 2025-06-26
Bump GLib version to 2.66

qarkai:
See https://github.com/BestImageViewer/geeqie/pull/1790#issuecomment-3009541624 and later for rationale.


caclark:
Implemented in commit a5234672f03693ef8c4fd445c17d0e63fa64cfb9

PR #1793, opened 2025-06-28
Use g_autolist(GtkTreePath)

qarkai:
Simplify related code.


caclark:
Implemented in commit 2b75c97b0ad570b4e2a9ef8ad2c52d7b21eb6405

PR #1795, opened 2025-06-29
Use g_autolist(WindowNames)

qarkai:


caclark:
Implemented in commit df124d92076ea1873420a3dbcbecd8d8cd89f2c4

PR #1798, opened 2025-07-04
per-image information text in the collection window

UweOhse:
In another project i create lists of similar images, sorted by a sharpness measure (it's a culling helper).
This value i want to show in the geeqie collection window.

This patch implements this by 
1) adding a new comment line type in the .gqv file:
#i some information
"/whereever/image.jpg"
#i another info
"/whereever/image.bla"

This information line is loaded and saved. In case there are multiple #i lines for one image the last one wins.

2) showing that information in the collection window, with a menu entry and a keyboard shortcut.

3) adding a new configuration option for persistance.

./scripts/test-all.sh ran through with this result:
```
Ok:                 1104
Expected Fail:      1   
Fail:               0   
Unexpected Pass:    0   
Skipped:            0   
Timeout:            0   
```

### DIFF
--- a/src/collect-io.cc
+++ b/src/collect-io.cc
@@ -199,6 +199,7 @@ static gboolean collection_load_private(CollectionData *cd, const gchar *path, C
 	guint flush = !!(flags & COLLECTION_LOAD_FLUSH);
 	guint append = !!(flags & COLLECTION_LOAD_APPEND);
 	guint only_geometry = !!(flags & COLLECTION_LOAD_GEOMETRY);
+	g_autofree gchar *infotext = nullptr;
 
 	if (!only_geometry)
 		{
@@ -247,7 +248,15 @@ static gboolean collection_load_private(CollectionData *cd, const gchar *path, C
 			/* Parse comments */
 			if (*p == '#')
 				{
+				if (strncmp(p, "#i ", 3 ) == 0)
+					{
+					g_free(infotext);
+					infotext = g_strdup(p+3);
+					gchar *q = strpbrk(infotext, "\r\n");
+					if (q) *q = 0;
+					}
 				if (!need_header) continue;
+
 				if (g_ascii_strncasecmp(p, GQ_COLLECTION_MARKER, GQ_COLLECTION_MARKER_LEN) == 0)
 					{
 					/* Looks like an official collection, allow unchecked input.
@@ -330,7 +339,11 @@ static gboolean collection_load_private(CollectionData *cd, const gchar *path, C
 		if (!flush)
 			changed |= collect_manager_process_action(entry, &filename);
 
-		if (filename[0] == G_DIR_SEPARATOR && collection_add_check(cd, file_data_new_simple(filename), FALSE, TRUE)) continue;
+		if (filename[0] == G_DIR_SEPARATOR && collection_add_check(cd, file_data_new_simple(filename), FALSE, TRUE, infotext)) 
+			{
+			g_clear_pointer(&infotext, g_free);
+			continue;
+			}
 
 		log_printf("Warning: Collection: %s Invalid file: %s", cd->name, filename);
 		DEBUG_1("collection invalid file: %s", filename);
@@ -376,7 +389,7 @@ static gboolean collection_load_private(CollectionData *cd, const gchar *path, C
 		gchar *buf = nullptr;
 		while (collect_manager_process_action(entry, &buf))
 			{
-			collection_add_check(cd, file_data_new_group(buf), FALSE, TRUE);
+			collection_add_check(cd, file_data_new_group(buf), FALSE, TRUE, nullptr);
 			changed = TRUE;
 			g_free(buf);
 			buf = nullptr;
@@ -531,6 +544,8 @@ static gboolean collection_save_private(CollectionData *cd, const gchar *path)
 	for (GList *work = cd->list; work; work = work->next)
 		{
 		auto ci = static_cast<CollectInfo *>(work->data);
+		if (ci->infotext && *ci->infotext)
+			g_string_append_printf(gstring, "#i %s\n", ci->infotext);
 
 		g_string_append_printf(gstring, "\"%s\"\n", ci->fd->path);
 		}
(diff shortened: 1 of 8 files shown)


src/collect-io.cc
```diff
@@ -247,7 +248,21 @@ static gboolean collection_load_private(CollectionData *cd, const gchar *path, C
 			/* Parse comments */
 			if (*p == '#')
 				{
+				if (strncmp(p, "#i ", 3 ) == 0)
+					{
+					gchar *q;
+					infotext = g_strdup(p+3);
+					for (q=infotext;*q; q++)
```
qarkai, line 255: Prefer https://en.cppreference.com/w/cpp/string/byte/strpbrk.html

src/collect-io.cc
```diff
@@ -199,6 +199,7 @@ static gboolean collection_load_private(CollectionData *cd, const gchar *path, C
 	guint flush = !!(flags & COLLECTION_LOAD_FLUSH);
 	guint append = !!(flags & COLLECTION_LOAD_APPEND);
 	guint only_geometry = !!(flags & COLLECTION_LOAD_GEOMETRY);
+	gchar *infotext = nullptr;
```
qarkai, line 202: Consider reducing scope and using `g_autofree`.

src/collect-table.cc
```diff
@@ -2423,21 +2447,31 @@ static void collection_table_cell_data_cb(GtkTreeViewColumn *, GtkCellRenderer *
...
 			{
-			display_text = g_strconcat(info->fd->name, "\n", star_rating, NULL);
+			display_text = g_strdup(info->fd->name);
 			}
-		else if (ct->show_text)
+
+		if (ct->show_stars && display_text)
 			{
-			display_text = g_strdup(info->fd->name);
+			gchar *t = g_strconcat(display_text,"\n",star_rating,NULL);
```
qarkai, line 2457: Please use `g_autofree` and `std::swap`. Also respect formatting, i.e. add space after commas.

src/collect.cc
```diff
@@ -122,6 +123,7 @@ void collection_info_free(CollectInfo *ci)
 
 	file_data_unref(ci->fd);
 	collection_info_free_thumb(ci);
+	if (ci->infotext) g_free(ci->infotext);
```
qarkai, line 126:
```suggestion
	g_free(ci->infotext);
```
Redundant check.

src/collect.h
```diff
@@ -107,8 +108,8 @@ void collection_randomize(CollectionData *cd);
 void collection_set_update_info_func(CollectionData *cd,
 				     void (*func)(CollectionData *, CollectInfo *, gpointer), gpointer data);
 
-gboolean collection_add(CollectionData *cd, FileData *fd, gboolean sorted);
-gboolean collection_add_check(CollectionData *cd, FileData *fd, gboolean sorted, gboolean must_exist);
+gboolean collection_add(CollectionData *cd, FileData *fd, gboolean sorted, gchar *infotext);
+gboolean collection_add_check(CollectionData *cd, FileData *fd, gboolean sorted, gboolean must_exist, gchar *infotext);
```
qarkai, line 112: `const gchar *infotext`

src/collect-table.cc
qarkai: On second thought consider using GString for `display_text` build.

src/collect-io.cc
UweOhse: It's in the right scope, it think - because the infotext is needed in the next loop iteration.
g_autofree: +

src/collect-table.cc
UweOhse: Switched to GString. Is my use of g_autoptr correct?

UweOhse:
> @UweOhse Thanks for the patch.
> 
> Unfortunately I do not understand how the `infotext` is input.

it's generated by an external tool creating the .gqv file. Right now it's something like
#i 1.23*, image 4/56 in group 879.
which just fits below the thumb (for me).

> The function `collection_add` is always called with the `comment` parameter as `nullptr`.

i renamed comment to infotext (that was a leftover).

i'm quite undecided whether to add a possibility to enter / edit it in geeqie. It somehow seems incomplete to not have it, but on the other hand i never needed it in 20 years.

I don't know.

caclark:
> i'm quite undecided whether to add a possibility to enter / edit it in geeqie. It somehow seems incomplete to not have it, but on the other hand i never needed it in 20 years.

Wait until someone complains. The complainer will have an idea about how they wish to make the input.
Until then it is not worth trying to predict what the requirements may be.

src/collect-io.cc
qarkai: I see, there are a lot of `break` s and `continue`s in loop. Parsing is a bit complex.

src/collect-io.cc
```diff
@@ -330,7 +339,12 @@ static gboolean collection_load_private(CollectionData *cd, const gchar *path, C
 		if (!flush)
 			changed |= collect_manager_process_action(entry, &filename);
 
-		if (filename[0] == G_DIR_SEPARATOR && collection_add_check(cd, file_data_new_simple(filename), FALSE, TRUE)) continue;
+		if (filename[0] == G_DIR_SEPARATOR && collection_add_check(cd, file_data_new_simple(filename), FALSE, TRUE, infotext)) 
+			{
+			if (infotext) g_free(infotext);
+			infotext = nullptr;
```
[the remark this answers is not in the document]
UweOhse: If i understood the documentation of g_clear_pointer correctly, then i'd need to write a wrapper function for g_free, which wouldn't improve readability in my eyes.
Would that really be better?

qarkai:
LGTM.
@UweOhse Could you squash commits after resolving all comments? "Fix" commits don't need to be in master branch.

UweOhse:
That's not what you wanted, right?
Would it be okay if i created  another branch with just the one commit and made a new pull request for that?

qarkai:
Not really. Try this

```sh
git switch collection-infotext
git reset --soft $(git merge-base master HEAD)
git commit -m "Allow to show a per-image information text in the collection window, loaded from and re-saved to the .gqv file"
git push --force-with-lease
```

UweOhse:
Thank you for your help!

caclark:
Implemented in commit 01d997edd9d55fd570a2650ad6d59933a5dd374e

caclark:
@UweOhse 
For information only: 593e6287ebd39477dfa0a52f6556f2c4114b8efb

caclark:
@UweOhse 
For information only: 5ad706a20a98571722ee529766967b944dc408c7

PR #1801, opened 2025-07-06
Drop redundant menu choice callbacks

qarkai:
No need to get combo value until search start.

Also drop unused `SearchData::search_exif` member.
Move `menu_choice_get_match_type()` to anonymous namespace.


caclark:
Implemented in commit 29474091910dac841b50b4bbb0391db40b139485

PR #1810, opened 2025-07-17
Drop read-only fullscreen.above option

qarkai:
Always had `FALSE` value.
Also drop remaining code for "above" button and simplify `fullscreen_prefs_selection_cb()`.


caclark:
I am confused, as always.
Commit a16635ca1d is also involved.

The documentation about keep-above:
https://docs.gtk.org/gtk3/method.Window.set_keep_above.html
_NET_WM_STATE_ABOVE in https://specifications.freedesktop.org/wm-spec/1.3/ar01s05.html

-----
I am using a two monitor setup running Wayland. I commented out line 452 in `fullscreen.cc`, so that keep-above is always set.
Run Geeqie, put a fullscreen image on one monitor, and drag a window from the other monitor onto the fullscreen image.
The window stays on top.

I do not see that this feature has any effect, although maybe under X11 things are different.

caclark:
> So let's remove option completely.

OK

caclark:
Implemented in commit dab8dc9faf057f810981a9e4221b49931bf18248

PR #1813, opened 2025-07-21
Drop FullScreenData::stop_data, use lambdas

qarkai:


caclark:
Implemented in commit bebbe165f5e1f67874056e2ba70b4d5da331a650

PR #1815, opened 2025-07-22
Add BarSort struct to deduplicate code

qarkai:
Move related enums to `BarSort` struct.
Remove redundant `SortData::help_button` member.


caclark:
Implemented in commit 3ae5f5902516c9a8226efe86fd254dafa4de1627

PR #1816, opened 2025-07-23
Fix memory leaks and simplify bar_sort_bookmark_select*() functions

qarkai:


caclark:
Implemented in commit 0c29671ff061a00d5ec548225d9068f47fbfb822

PR #1818, opened 2025-07-23
Deduplicate file_util_delete_notify_done() calls

qarkai:


qarkai:
Looks like `ConfOptions::file_ops.safe_delete_enable` could be converted to `file_util_delete()`/`file_util_delete_notify_done()` parameter.
It is always set before these functions call and used only inside `file_util_delete_full()`.

No need to place it into config file.

If there are no objections I'll change it in next PR.

caclark:
Implemented in commit 7291c89a99818ebd25cd97beb36f72edbe4c0f64

PR #1825, opened 2025-07-26
Deduplicate layout menu rating callbacks

qarkai:


caclark:
Implemented in commit a47785742f6fba45db972e5bd96fe08851d7c2ef

PR #1832, opened 2025-07-27
Deduplicate image set pixbuf renderer post process function

qarkai:


caclark:
Implemented in commit a9a4397a29959c23eeb5db7d1878c0e3bfd60179

PR #1833, opened 2025-07-27
Drop PixbufRenderer callbacks user data

qarkai:


caclark:
Implemented in commit 35b579fe71ea13491581a2ed638366e83f65d481

PR #1835, opened 2025-08-01
Use g_format_size_full() for text_from_size_abrev()

qarkai:


qarkai:
@caclark
From failed `FileDataTest.text_from_size_abrev_test`
```
Expected equality of these values:
  test_case.second
    Which is: "1.0 KiB"
  std::string(generated)
    Which is: "1.0\xC2\xA0KiB"
    As Text: "1.0 KiB"
```
Apparently GLib uses `NO-BREAK SPACE` `\xC2\xA0` instead of space.

I suggest to drop `FileDataTest.text_from_size_abrev_test` instead of fixing test cases. Since now we actually test `g_format_size_full`.

caclark:
@qarkai 

OK

PR #1836, opened 2025-08-02
Drop CollectionData::info_updated_data, use lambdas

qarkai:


caclark:
Implemented in commit d8cc93a60f4fae0dc40bad3d0898a9dd9b5f3d8b

PR #1846, opened 2025-08-08
Fix memory leaks after gtk_action_group_list_actions() calls

qarkai:
Optimize `menu_item_add_main_window_accelerator()` a bit by checking `accel_key` earlier.


caclark:
Implemented in commit 31b820e2b001c46f95a460b4f3bb1f4e66f6a5e4

PR #1855, opened 2025-08-13
Fix memory leaks in bookmark_string() calls

qarkai:


caclark:
Implemented in commit fb74b9f7709125c1bdad844caee70cdcea4e8c34

PR #1856, opened 2025-08-14
Drop constant parameters from submenu_add_sort(), inline submenu_add_dir_sort()

qarkai:
Move adding `SORT_PATH` in collection menu to collect-table.
Simplify `vd_pop_menu_sort_cb()`.


caclark:
Implemented in commit 58b5706f7fdec94a29c4282af02e35a14531dc09

PR #1857, opened 2025-08-15
Inline submenu_add_sort_item()

qarkai:


caclark:
Implemented in commit dcac870b90597ca9b412db9db2c3c0a2abf62630

PR #1859, opened 2025-08-16
Allow nullptr for item_data in menu_item_add_radio()

qarkai:
to simplify callbacks.


caclark:
Implemented in commit e8acf6f73fe97a1348d98cd021d6806f49f53592

PR #1867, opened 2025-08-17
ExifUnitType is used in exif.cc only

qarkai:


caclark:
Implemented in commit 99ed926b83d32288437d8ecb38b805da82987e37

PR #1873, opened 2025-08-20
Drop redundant BarData::add_button

qarkai:


caclark:
Implemented in commit 868e57d0d947a417c011b053b04d478ee9520268

PR #1877, opened 2025-08-23
Implement filechooser

caclark:
This is a request for comment, rather than a potential commit.

1. All folder/file dialogs are via gtk_file_chooser(). This provides a consistent user interface, and the dialog itself has advantages.

2. The dialog is called from a single location, and the interface is via a simple structure, except for the With Rename action

3. Previews of folders/some images/pdf/icc/some text files.

4. A history combo is included where sensible.

Regression:

The copy/move dialog had a "dot" as a shortcut to the current folder. The new dialog also has a shortcut to the current folder, but its name is the final folder name in the path - there is no way to change it to be a "dot". It is therefore not easy to see which the current folder shortcut is.


src/utilops.h
```diff
@@ -44,10 +44,12 @@ GenericDialog *file_util_gen_dlg(const gchar *title,
 				 const gchar *role,
 				 GtkWidget *parent, gint auto_close,
 				 void (*cancel_cb)(GenericDialog *, gpointer), gpointer data);
-FileDialog *file_util_file_dlg(const gchar *title,
+GtkWidget *file_util_file_dlg(const gchar *title,
 			       const gchar *role,
 			       GtkWidget *parent,
-			       void (*cancel_cb)(FileDialog *, gpointer), gpointer data);
+			       void (*cancel_cb)(FileDialog *, gpointer),
```
qarkai, line 50: `FileDialog` and related functions are not used anymore and could be removed.

Also whole `ui-pathsel` module becomes unused.

src/ui-file-chooser.cc
```diff
@@ -0,0 +1,662 @@
...
+
+			gtk_file_filter_add_pattern(sub_filter, ext_pattern);
+			}
+		gtk_file_chooser_add_filter(GTK_FILE_CHOOSER(dialog), sub_filter);
+		gtk_file_chooser_set_filter(GTK_FILE_CHOOSER(dialog), sub_filter);
+		}
+
+	g_signal_connect(dialog, "response", G_CALLBACK(fcdd->response_callback), fcdd->data);
+
+	/* It is expected that extra_widget contains only a single widget - i.e. that
```
qarkai, line 552: Is it possible to use container with multiple children?

src/bar-sort.cc
```diff
@@ -477,33 +472,26 @@ static void bar_sort_add_cb(GtkWidget *, gpointer data)
...
-		g_signal_connect(dialog, "response", G_CALLBACK(new_collection_file_response_cb), sd);
-
-		gtk_window_present(GTK_WINDOW(dialog));
+		g_autoptr(FileChooserDialogData) fcdd = g_new0(FileChooserDialogData, 1);
+
+		fcdd->action = GTK_FILE_CHOOSER_ACTION_SAVE;
+		fcdd->accept_text = _("Save");
+		fcdd->data = sd;
+		fcdd->do_overwrite_confirmation = TRUE;
+		fcdd->entry_text = _("Optional name...");
```
qarkai, line 481: Both `entry_text` and `history_key` are set.

qarkai (commented):
Sorry for the late response. I believe these issues should be addressed in future PRs.

PR #1880, opened 2025-08-26
Drop dead code

qarkai:
GLib required version is 2.66.


caclark:
Implemented in commit 7174b038ee057e59e75c1a3b92523bf4bce1de1a

PR #1886, opened 2025-08-31
Deduplicate more command line callbacks

qarkai:
Fix `cache-render-shared-recurse` option handling.


caclark:
Implemented in commit 8399276f85e1e81c07493a116d12c1b3a089c54d

PR #1887, opened 2025-08-31
Make pan search callbacks local static

qarkai:


caclark:
Implemented in commit 44887555ce851ae74572d4b137a12863eef6de90

PR #1890, opened 2025-08-31
Drop unused ui-pathsel

qarkai:


caclark:
Implemented in commit 1db6dbdfe1215dee4a4af8f54c149ef5f9c83569

PR #1893, opened 2025-09-01
Fix leaks

caclark:
Another try.


caclark:
@qarkai 
Thanks for the help

PR #1896, opened 2025-09-04
Drop read-only TabCompData::dialog member

qarkai:
Also drop unused TabCompData::response_callback.


caclark:
Implemented in commit f55d398af7a4862a61ed7c36273045f263317e26

PR #1897, opened 2025-09-04
Simplify setting tab completion callbacks

qarkai:


caclark:
Implemented in commit 5502e9ac9dc425782ca932be4b3eec5efbef0020

Issue #1894, opened 2025-09-02
Fails to build from source in musl


icp1994:
**Setup (please complete the following information):**
 - Distribution:Void Linux
 - Distribution release: rolling
 - Geeqie version [geeqie --version]: 2.6.1

**Describe the bug**
Fails to compile in musl arches

You can see the full build log at https://github.com/void-linux/void-packages/actions/runs/15410493705/job/43557302478

My understanding is that [this block](https://github.com/BestImageViewer/geeqie/blob/v2.6.1/src/collect-io.cc#L158-L173) targetomg non-glibc platforms is primarily compatible with BSD systems and not musl libc, which doesn't define `MNT_NOWAIT` and uses `getmntent` like glibc.

PR #1898, opened 2025-09-05
Fix build issue using musl reported in #1894

EvTheFuture:
The commit 16fb13d doesn't resolve build issues on Alpine Linux This patch has been tested with success on Alpine Linux


xsdg:
I don't like this change, though it's good to know that this is enough to get it compiling.  glibc and musl aren't the only libc implementations out there, and if musl doesn't want to make itself easy to identify, that shouldn't cause us to potentially break other non-glibc libc implementations that aren't musl.

I would much prefer that we use a detection-based approach like this one: \
https://stackoverflow.com/a/78062291 \
And then change the preprocessor directives to something like `#if defined(__GLIBC__) || defined (__MUSL__)`

Tagging @qarkai since this is related to header organization

lotheac:
@qarkai another alpine user here. build output at current master 5502e9ac9dc425782ca932be4b3eec5efbef0020 is:
```
❯ ninja
[1/2] Compiling C++ object src/geeqie.p/collect-io.cc.o
ninja: job failed: c++ -Isrc/geeqie.p -Isrc -I../src -I. -I.. -Ipo -Isrc/authors -Isrc/icons -Isrc/ui -I/usr/include/gtk-3.0 -I/usr/include/pango-1.0 -I/usr/include/cairo -I/usr/include/gdk-pixbuf-2.0 -I/usr/include/webp -I/usr/include/at-spi2-atk/2.0 -I/usr/include/at-spi-2.0 -I/usr/include/atk-1.0 -I/usr/include/dbus-1.0 -I/usr/lib/dbus-1.0/include -I/usr/include/libdrm -I/usr/include/fribidi -I/usr/include/pixman-1 -I/usr/include/harfbuzz -I/usr/include/freetype2 -I/usr/include/libpng16 -I/usr/include/gio-unix-2.0 -I/usr/include/glib-2.0 -I/usr/lib/glib-2.0/include -I/usr/include/libmount -I/usr/include/blkid -fdiagnostics-color=always -D_GLIBCXX_ASSERTIONS=1 -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -Wpedantic -std=c++14 -O2 -g -include debug.h -pthread -MD -MQ src/geeqie.p/collect-io.cc.o -MF src/geeqie.p/collect-io.cc.o.d -o src/geeqie.p/collect-io.cc.o -c ../src/collect-io.cc
../src/collect-io.cc: In function 'bool {anonymous}::is_file_on_mounted_drive(const gchar*)':
../src/collect-io.cc:159:46: error: 'MNT_NOWAIT' was not declared in this scope
  159 |         int num_mounts = getmntinfo(&mounts, MNT_NOWAIT);
      |                                              ^~~~~~~~~~
../src/collect-io.cc:159:26: error: 'getmntinfo' was not declared in this scope
  159 |         int num_mounts = getmntinfo(&mounts, MNT_NOWAIT);
      |                          ^~~~~~~~~~
../src/collect-io.cc:170:44: error: invalid use of incomplete type 'struct {anonymous}::is_file_on_mounted_drive(const gchar*)::statfs'
  170 |                 if (is_file_in_dir(mounts[i].f_mntonname))
      |                                            ^
../src/collect-io.cc:158:16: note: forward declaration of 'struct {anonymous}::is_file_on_mounted_drive(const gchar*)::statfs'
  158 |         struct statfs* mounts;
      |                ^~~~~~
../src/collect-io.cc:170:44: error: invalid use of incomplete type 'struct {anonymous}::is_file_on_mounted_drive(const gchar*)::statfs'
  170 |                 if (is_file_in_dir(mounts[i].f_mntonname))
      |                                    ~~~~~~~~^
../src/collect-io.cc:158:16: note: forward declaration of 'struct {anonymous}::is_file_on_mounted_drive(const gchar*)::statfs'
  158 |         struct statfs* mounts;
      |                ^~~~~~
ninja: subcommand failed
```

this is because musl does not define any macros that allow you to detect it; upstream's stance is that you should detect _features_, not libc's. and so, because `defined(__MUSL__)` is not hit, it tries to compile the getmntinfo variant, which doesn't work on Linux.

I can see commit 4709e028015630193d5a388651fe010dd42fc89f changed the original `__linux__` check to `__GLIBC__`, but that was also not right. `mntent.h` (`setmntent` etc.) is an API that is available on all Linux libc's that I could find:

* glibc (obviously)
* [musl](https://git.musl-libc.org/cgit/musl/tree/include/mntent.h)
* [uClibc](https://git.uclibc.org/uClibc/tree/include/mntent.h)
* [bionic](https://github.com/aosp-mirror/platform_bionic/blob/main/libc/include/mntent.h) (despite bionic being based on OpenBSD libc, which does _not_ have mntent.

therefore, my suggestions are:

* change the check to `defined(__linux__) || defined(__GLIBC__)` to accommodate both musl on Linux and glibc on non-Linux, OR
* use feature detection: ie. add "configure-time" compile tests to see whether code using `setmntent` or `getmntinfo` can be compiled, store the result in a generated header file, and use that in the code

EvTheFuture:
Updated the path to detect if 'mntent.h' exists and if so sets the define HAVE_MNTENT_H

Then in collect-io.cc:

#if defined(HAVE_MNTENT_H) || defined(__GLIBC__)

Would this be a suitable solution? This shouldn't break building on currently working systems, right?

src/collect-io.cc
```diff
@@ -29,7 +29,7 @@
 #include <gdk/gdk.h>
 #include <glib-object.h>
 
-#if defined(__GLIBC__)
+#if defined(HAVE_MNTENT_H) || defined(__GLIBC__)
```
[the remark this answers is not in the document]
EvTheFuture: Yeah, you're correct.

meson.build
```diff
@@ -142,6 +142,13 @@ summary({'gq_appdir': gq_appdir,
 conf_data = configuration_data()
 conf_data.set('DEBUG', debug)
 
+
+# Detect if mntent.h is available
+has_mntent_h = cc.has_header('mntent.h')
+if has_mntent_h
+    add_project_arguments('-DHAVE_MNTENT_H', language: ['c', 'cpp'])
```
qarkai, line 149: Please use `conf_data.set()` to unify project configuration.

EvTheFuture: Done

qarkai:
LGTM

PR #1908, opened 2025-09-13
Move setting TabCompData fd_* members to tab_completion_add_select_button()

qarkai:


caclark:
Implemented in commit fe95845f1b9bacadf1db1da57cbcc4b72b9fc495

PR #1915, opened 2025-09-18
Drop TAB_COMPLETION_ENABLE_POPUP_MENU

qarkai:
It is always defined.


caclark:
Implemented in commit f11493a32f36761fde79c4eb9cdb4e7a369b354c

PR #1916, opened 2025-09-19
Add TabCompPrefix::match() method to encapsulate search logic

qarkai:


caclark:
Implemented in commit 5c44b6c75a6ae0660621aba9b24a93355d57fc77

PR #1917, opened 2025-09-20
Drop TabCompData data members, use lambdas

qarkai:


caclark:
Implemented in commit 946fa90bee710f28fbd7f3ef3d54c3a3eacca2e2

PR #1927, opened 2025-09-28
Drop SlideShowData::stop_data, use lambdas

qarkai:


caclark:
Implemented in commit ff755388726f7fdfecf452bc3896427191e318ac

PR #1938, opened 2025-10-04
Return value from collection_ref() to simplify assignment

qarkai:


caclark:
Implemented in commit 4e8d0d9536bb6a549ccec28a1f3ca5ef1766ddc5

PR #1942, opened 2025-10-05
Make bytes parameter optional

qarkai:


caclark:
Implemented in commit 2af9d619048faabfe126cf08563d95179118ebcd

PR #1943, opened 2025-10-05
Optimize collection_table_update_status() a bit

qarkai:
Calculate selection list size only if need.


caclark:
Implemented in commit caa1a1aa099d01f2b90b6181ebac966e87d18eab

PR #1944, opened 2025-10-06
Simplify image_get_collection()

qarkai:
Search `imd->collection_info` only if need.


caclark:
Implemented in commit 7d7baf052a4b103579c052bba466dcf9cbc7eb34

PR #1946, opened 2025-10-07
Fix build on NetBSD with gcc 14.

0-wiz-0:
error: 'abs' is not a member of 'std'; did you mean 'abs'?

```
../src/command-line-handling.cc: In function 'void {anonymous}::gq_get_rectangle(GtkApplication*, GApplicationCommandLine*, GVariantDict*, GList*)':
../src/command-line-handling.cc:818:65: error: 'abs' is not a member of 'std'; did you mean 'abs'?
  818 |                                                            std::abs(x1 - x2),
      |                                                                 ^~~
```

### DIFF
--- a/src/command-line-handling.cc
+++ b/src/command-line-handling.cc
@@ -20,6 +20,7 @@
 
 #include "command-line-handling.h"
 
+#include <cmath>
 #include <cstdlib>
 #include <vector>
 


qarkai:
> error: 'abs' is not a member of 'std'; did you mean 'abs'?
> 
> ```
> ../src/command-line-handling.cc: In function 'void {anonymous}::gq_get_rectangle(GtkApplication*, GApplicationCommandLine*, GVariantDict*, GList*)':
> ../src/command-line-handling.cc:818:65: error: 'abs' is not a member of 'std'; did you mean 'abs'?
>   818 |                                                            std::abs(x1 - x2),
>       |                                                                 ^~~
> ```

Looks like `#include <cmath>` is missing.

qarkai:
@0-wiz-0 there are a [number of PRs](https://github.com/BestImageViewer/geeqie/issues?q=is%3Apr%20state%3Aclosed%20netbsd) fixing build on NetBSD. Is NetBSD very different from FreeBSD?

@caclark Looks like check compile on NetBSD https://github.com/vmactions/netbsd-vm should be added.

0-wiz-0:
Thanks for the suggestion, using `cmath` works as well. I've updated the PR.
NetBSD and FreeBSD diverged in the 1990s, so you should think of them as separate operating systems (coming from the same base).

PR #1968, opened 2025-10-17
Fix memory leaks in toolbarlist_populate()

qarkai:
Simplify related code.


caclark:
Implemented in commit 04d9006f1073d69de0fe740843e66ffb66f42733

PR #1973, opened 2025-10-19
Make layout_key_press_cb() local static

qarkai:
Use `layout_keyboard_init()`.
Simplify `layout_key_press_cb()` a bit.


caclark:
Implemented in commit 18ebafc07f0cd86f02620cb1e139105df31bbb4c

PR #1976, opened 2025-10-19
Simplify vf_file_filter_save_cb() a bit

qarkai:
Fix and utilize first parameter.
Remove redundant memory allocation.
Check if `entry_text` is empty once.


caclark:
Implemented in commit a7d268e87f28dd8ce740a802b6e2aab8ede0108e

PR #1982, opened 2025-10-21
Fix memory leak in create_pdf_preview()

qarkai:
Simplify a bit.


qarkai:
`PopplerDocument` supports `g_autoptr` since 25.06. Apparently `PopplerPage` too.

PR #1984, opened 2025-10-22
Move MetadataFormat enum to metadata.h

qarkai:


caclark:
Implemented in commit a753e968cdc285eace7cf245149327b29e24cffd

PR #1991, opened 2025-10-25
Drop unused typedefs.h

qarkai:


caclark:
Implemented in commit 8d26e6ef209b2fe49db502ff6bb5513060a7ac63

PR #1995, opened 2025-10-28
Reduce code duplication in filecache

qarkai:


caclark:
Implemented in commit 28c52e8a0133016d2140eea0240c99449786f1bb

PR #1996, opened 2025-10-29
Simplify static exif_cache initialization

qarkai:


caclark:
Implemented in commit aedbe270acb3f0bea1de026e1192f8d00719bec6

PR #2004, opened 2025-11-02
Move pref_link_sensitivity() to anonymous namespace

qarkai:
Replace deprecated `state_changed` signal with `state-flags-changed` signal.


caclark:
Implemented in commit d10aa7a06d3ec9675f834229125ae1c544f26d91

PR #2008, opened 2025-11-04
Stop overwriting prepared image data

### LINKED ISSUE #1921: PNGs not rendering on new arch install
 - Distribution: Arch
 - Distribution release: Latest
 - Geeqie version [geeqie --version]: Geeqie 2.6.1 GTK3

**Describe the bug**
All PNG files are being displayed as either fully black or fully transparent images, the images work fine in other viewers and JPG and GIF seem to work fine.

xsdg:
As mentioned in the bugs referenced, clearing the pixbuf in the prepared signal is not a valid thing to do.

See: https://gitlab.gnome.org/GNOME/gdk-pixbuf/-/commit/75ac8deaf04f362efa5fc0ace51031350e70e8be
Closes: https://github.com/BestImageViewer/geeqie/issues/1921

[Added comment by @xsdg]
This is a version of #1928 with the "Don't remove PATH from tests" commit dropped.  Consequently, this should be considered a **partial fix** that will make the same code change as the original, but where the geeqie functional tests will still be broken with glycin (which is already the case, given the original bug)

### DIFF
--- a/src/image-load.cc
+++ b/src/image-load.cc
@@ -528,32 +528,8 @@ static void image_loader_area_updated_cb(gpointer,
 	g_mutex_unlock(il->data_mutex);
 }
 
-static void image_loader_area_prepared_cb(gpointer, gpointer data)
+static void image_loader_area_prepared_cb(gpointer, gpointer)
 {
-	auto il = static_cast<ImageLoader *>(data);
-	GdkPixbuf *pb;
-	guchar *pix;
-	size_t h;
-	size_t rs;
-
-	/* a workaround for
-	   https://bugzilla.gnome.org/show_bug.cgi?id=547669
-	   https://bugzilla.gnome.org/show_bug.cgi?id=589334
-	*/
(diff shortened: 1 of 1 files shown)


src/image-load.cc
```diff
@@ -528,32 +528,8 @@ static void image_loader_area_updated_cb(gpointer,
 	g_mutex_unlock(il->data_mutex);
 }
 
-static void image_loader_area_prepared_cb(gpointer, gpointer data)
```
qarkai: Are we sure it's the proper way fixing GdkPixbufLoader's issue?
Maybe it's enough to remove this handler in ImageLoaderGdk and ImageLoaderSvgz?

qarkai: This callback is still used by jpeg and tiff loaders which do not rely on GdkPixbufLoader.
If we keep it then workaround for svg and xpm could be removed.

caclark:
Implemented in commit 7344b441cf4b62fb5d472fcb9f9ccb1b625c8971

PR #2009, opened 2025-11-05
Set result value in pref_color_button_new()

qarkai:


caclark:
Implemented in commit d62bf36efe8570fc8e7e54ff2fe7cc96c350c6e4

PR #2014, opened 2025-11-09
Move shift_color() to misc and simplify a bit

qarkai:


caclark:
Implemented in commit aa58a75438b5893d2097ea2b65c259c0f28335da

PR #2020, opened 2025-11-09
widget_auto_scroll_start() return value is not used

qarkai:


caclark:
Implemented in commit 9b9faf16e4179ae853f4a94f2cac9290e9495ac9

PR #2023, opened 2025-11-10
Fix memory leak and simplify a bit

qarkai:


caclark:
Implemented in commit e07ced9957d67b88b107936e23c2754222656cff

PR #2024, opened 2025-11-12
Drop ImageLoaderBackend::AreaPreparedCb

qarkai:


qarkai:
@caclark
I have noticed differences in parameter types of image-load signals/callbacks.

1. [ImageLoaderBackend::AreaUpdatedCb](https://github.com/BestImageViewer/geeqie/blob/f6c7aba626ae762dec56173144be62914423a063/src/image-load.h#L40) is supposed to be callback for `GdkPixbufLoader::area-updated` signal, see https://github.com/BestImageViewer/geeqie/blob/f6c7aba626ae762dec56173144be62914423a063/src/image-load-gdk.cc#L84 But `ImageLoaderBackend::AreaUpdatedCb` uses `guint` for updated area parameters while [area_updated](https://docs.gtk.org/gdk-pixbuf/signal.PixbufLoader.area-updated.html) declaration expects `gint`.
2. Similar with [ImageLoader::area_ready](https://github.com/BestImageViewer/geeqie/blob/f6c7aba626ae762dec56173144be62914423a063/src/image-load.cc#L179) signal.  It expects `gint`s but callback https://github.com/BestImageViewer/geeqie/blob/f6c7aba626ae762dec56173144be62914423a063/src/image.cc#L847 uses `guint`s.

Do we really need unsigned types there? In other words are there images with more than `INT_MAX` (2147483647 usually) width and/or height in pixels?

I suggest to fix `ImageLoaderBackend::AreaUpdatedCb`, `image_load_area_cb()` parameter types.
Also we can change `ImageLoader::area_ready` parameters from four `int` to `GdkRectangle` to simplify code a bit.

PR #2030, opened 2025-11-15
Fix ImageLoader::area-ready signal

qarkai:
See https://github.com/BestImageViewer/geeqie/pull/2024#issuecomment-3533351724


caclark:
Implemented in commit 477c5acc8c895a5d538512edbd44cebb30564bab

PR #2031, opened 2025-11-15
Use GdkRectangle for ImageLoader::area-ready signal

qarkai:


caclark:
Implemented in commit 52323bdc296d92f1bb3994c6c3f990042deaccdc

PR #2035, opened 2025-11-19
Add FileData::supports_exif_orientation() to reduce duplication

qarkai:
Simplify related code a bit.


qarkai:
@xsdg I'm not sure if `FileData` is right class for this method but could not find more suitable place.

PR #2037, opened 2025-11-23
Use maybe_unused attribute

qarkai:


caclark:
Implemented in commit 3b731128ed4c8bfccfec7f2f7af66e9a34532534

PR #2039, opened 2025-11-25
Use std::clamp and simplify a bit

qarkai:


caclark:
Implemented in commit 80cc62bc7340e8e37ece51d8fca50af1cc55eca2

qarkai:
Actually `std::clamp()` revealed `CLAMP` usage bugs.
`CLAMP(value, low, high)` [documentation](https://docs.gtk.org/glib/macros.html#maths:~:text=CLAMP(value%2C%20low%2C%20high)) says `If low is greater than high the result is undefined`  but checked nothing and result according to macro definition
```
(((x) > (high)) ? (high) : (((x) < (low)) ? (low) : (x)))
```
is equal to `high` if `x` is more than `low`.

`std::clamp()` (at least in GCC) has run-time check `!(__hi < __lo)` which produces run-time errors. So in order to fix them we have to check boundaries before calling `std::clamp()`. This check was required for `CLAMP` too.

Upd. I'll try to fix this later.

PR #2040, opened 2025-11-25
Simplify sig_handler_cb() a bit

qarkai:


qarkai:
@caclark How do I fix 'Untranslated text' check fail?

PR #2042, opened 2025-11-27
Use GqSize for thumb_size_list

qarkai:


caclark:
Implemented in commit 3d84ab3478f3fefc6dd8acd526956400ec391c49

PR #2045, opened 2025-11-29
Use GdkPoint in some PixbufRenderer functions

qarkai:


caclark:
Implemented in commit dbfcaa79effef0c12d7e26144f68fde212553f76

PR #2048, opened 2025-11-30
Return value from image_get_rectangle()

qarkai:
instead of out parameters.


caclark:
Implemented in commit 0c8a7e900bea308e320e59ddc14058a7deacb3f4

PR #2052, opened 2025-12-08
Replace deprecated gdk grab functions

qarkai:


caclark:
@qarkai 

The effect happens only when fullscreen is started by double-click. By any other method there is no problem.

PR #2063, opened 2025-12-27
Fix ColorMan memory leak

qarkai:


caclark:
Implemented in commit 4e4ee4b1516b1e2493896f9d49990d3b6b5a2885

PR #2089, opened 2026-01-11
Move ExifFormatType to exif-int.h

qarkai:
Fix getting undefined tag in exiv2.


caclark:
Implemented in commit a92906a69e010d6fe71c1f4151bd5e6252eae729

PR #2090, opened 2026-01-11
Use bool for ExifRational sign

qarkai:


caclark:
Implemented in commit e635ed4ab3e31fcd2136400e05e02f2422f57b17

PR #2092, opened 2026-01-13
Replace G_N_ELEMENTS with std::size

qarkai:


caclark:
Implemented in commit d11447a0e28c2dd0a57022203cd3ae021a9c9c0e

Issue #2086, opened 2026-01-08
Build failure - unchecked version dependency for pango


ljb279:
Building geeqie-2.6.1 fails on my Linux system with:

  ./src/print.cc:517:27: error: ‘PangoFontDescription_autoptr’ was not
   declared in this scope; did you mean ‘PangoFontDescription’?

I tracked this down and the auto-cleanup pointer for PangoFontDescription was added to pango in version 1.49.4. My distribution ships 1.48.11 so that feature is missing. I realize that my distribution tends to stick with old
versions of libraries. Still, please consider adding a dependency check in meson.build for pango version >=1.49.4.

caclark:
@qarkai 

What ever you think best...

Understanding versioning is, for me, a problem. The check-builds run on the latest LTS - 24.04, which uses glib 2.8 and pango 1.52.

LTS 26.04 will be arriving soon. Should Geeqie be updated to follow?
And what about anything other than Ubuntu.

PR #2095, opened 2026-01-16
Add Pango version check

qarkai:
See #2086


caclark:
Implemented in commit 34730a59621926fb7c7776bd445ba2dd64265e2b

PR #2097, opened 2026-01-16
Move exif-common static functions to anonymous namespace

qarkai:
Make exif_get_integer() and exif_get_rational() local.

### DIFF
--- a/src/exif-common.cc
+++ b/src/exif-common.cc
@@ -57,6 +57,20 @@ struct ExifFormattedText
 	gchar *(*build_func)(ExifData *exif);
 };
 
+gint exif_get_integer(ExifData *exif, const gchar *key, gint *value)
+{
+	ExifItem *item = exif_get_item(exif, key);
+
+	return exif_item_get_integer(item, value);
+}
+
+ExifRational *exif_get_rational(ExifData *exif, const gchar *key, bool *sign = nullptr)
+{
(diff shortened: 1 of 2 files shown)


caclark:
@qarkai 

What is your opinion of this section?
https://llvm.org/docs/CodingStandards.html#restrict-visibility

I agree with this statement:
`it is easy to see if it is marked static, but seeing if it is in an anonymous namespace requires scanning a big chunk of the file.`

(but other suggestions there I do not agree with)

qarkai:
Of course, `static` immediately indicates that a function or variable is "local." The main question is: why would anyone need this information? As LLVM says, reduced scope helps the compiler perform optimizations. I see no reason to know the difference. And if you ever need to, the IDE can show functions in anonymous namespaces.
How it looks in Qt Creator for example
qt-creator

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf22-use-an-unnamed-anonymous-namespace-for-all-internalnon-exported-entities recommends using namespaces.

In addition, there are public functions that are only used in the same file where they are defined.

Upd. @caclark If you have any reason to use `static`, I'm fine with it.

caclark:
OK

This was showing up when I run tests locally.

I will put an exclude for `llvm-prefer-static-over-anonymous-namespace` in .clang-tidy.

PR #2098, opened 2026-01-16
Return optional from exif_get_integer() and related

qarkai:


caclark:
Implemented in commit b26f6ea3f31a13215de7efb4821cf8940763361f

PR #2101, opened 2026-01-18
Use string_view for jpeg_segment_find() magic parameter

qarkai:


caclark:
Implemented in commit 1f9e21a1c484b4bea886f0cf344645a80ca65459

PR #2103, opened 2026-01-18
Split some functions in print, simplify a bit

qarkai:


caclark:
Implemented in commit 34d2d917a1a1a458bed411db1a538a5bee6c6248

PR #2108, opened 2026-01-20
Fix clang-tidy readability-math-missing-parentheses warning

qarkai:
#2107 follow-up


caclark:
Implemented in commit c408d0221d9123676c0f0ee13b1063abb5e95559

PR #2127, opened 2026-02-01
Simplify RendererTiles overlay clear

qarkai:
Inline `rt_overlay_free()` and `rt_overlay_list_clear()`.
Use proper unref function for `RendererTiles::overlay_buffer` in `renderer_free()`.


caclark:
Implemented in commit 3c41dd55fc9ef6414410dc180f2aa2a1fcd6d321

PR #2131, opened 2026-02-01
Convert ImageRenderType to enum class TileRender

qarkai:


caclark:
Implemented in commit b1d7fdda6b149eef56f5e2093fc7fc325fbe7c32

PR #2140, opened 2026-02-02
Fix warnings found by clang-tidy's bugprone-reserved-identifier

### LINKED ISSUE #2138: reserved identifier violation
I would like to point out that identifiers like “[`__GLUA_H`](https://github.com/BestImageViewer/geeqie/blob/3e489b2b545ff20fb20a21d128911a64f50d32ad/src/glua.h#L21-L22 "Update candidate")” and “[`_MAIN_DEFINES_H`](https://github.com/BestImageViewer/geeqie/blob/3e489b2b545ff20fb20a21d128911a64f50d32ad/src/main-defines.h#L22-L23 "Another update candidate")” [do not fit](https://wiki.sei.cmu.edu/confluence/display/cplusplus/DCL51-CPP.+Do+not+declare+or+define+a+reserved+identifier#DCL51CPP.Donotdeclareordefineareservedidentifier-NoncompliantCodeExample%28HeaderGuard%29 "Do not declare an identifier which is reserved for the compiler implementation.") to the expected naming convention of the C++ language standard.
Would you like to adjust your selection for unique names?

qarkai:
Can't enable `bugprone-reserved-identifier` in `.clang-tidy` since it alerts on GLib macros.

Fix #2138.

### DIFF
--- a/CODING.md
+++ b/CODING.md
@@ -183,7 +183,7 @@ Pragma: (Indentation 2 spaces)
 Headers:
 
 ```c
-#ifndef _FILENAME_H
+#ifndef FILENAME_H
 ```
 
 Use [Names and Order of Includes](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes) for headers include order.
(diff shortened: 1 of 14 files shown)


caclark:
Implemented in commit 84fe45ede428f749ba2b1f52a5dce6195c6cbb19

PR #2147, opened 2026-02-06
Fix memory leak in layout_editors_reload_idle_cb()

qarkai:
No need to create (and leak) complex widget. List of actions from current LayoutWindow is enough.


caclark:
Implemented in commit 9d4f8f05645ba4790de54a7c6a7134d29b431ab4

PR #2151, opened 2026-02-07
Replace deprecated gdk_cairo_create()

qarkai:
As suggested in https://docs.gtk.org/gdk3/func.cairo_create.html

Not sure if this is proper way but overlay works for me.


caclark:
Implemented in commit 63d6b9c0edb6ca9756e3027a0675ded6f49b3c9e

PR #2152, opened 2026-02-09
Return vector<int> from layout_selection_list_by_index()

qarkai:
To avoid manual memory management and int->pointer->int conversions.


caclark:
Implemented in commit e015a3875be77b6b986b292346383fc7d7218c92

PR #2154, opened 2026-02-12
Use vector for LayoutConfig::style_widgets

qarkai:
Simplify destroy callback.


caclark:
Implemented in commit 6e4a3770b75f4cfcd437f8427ccd00cdd9328e2e

PR #2163, opened 2026-02-14
Use LayoutLocation forward declaration

qarkai:


caclark:
Implemented in commit a346f41907aaade05e588da703d870bc6b10c5e4

PR #2164, opened 2026-02-14
Move functions from misc to where they used

qarkai:
Drop duplicated declaration of `convert_gdkcolor_to_gdkrgba()`.


caclark:
Implemented in commit 6febb76b6f204fbadfd8619c5764c6ceb4ee99b9

PR #2175, opened 2026-02-16
Simplify key_pressed() callback a bit

qarkai:


caclark:
Implemented in commit dd97fa0e4c1b0f0f237c3744f208363130eacd56

PR #2186, opened 2026-02-23
Fix copy image overlay options

qarkai:


caclark:
Implemented in commit 7f43415639172ae8733d7412e30d284026c58859

PR #2196, opened 2026-02-28
Extract setup default bookmarks to ui-bookmark

qarkai:


caclark:
Implemented in commit 6da54f7819014a011827b595cc25407904127d43

PR #2210, opened 2026-03-05
Use string for members in bookmark structs

qarkai:
Rename BookButtonData::key to text.


caclark:
Implemented in commit 94eeadec729ab8a1513ce24eae1f701f45b571b0

PR #2245, opened 2026-03-18
Use string for PanTextAlignment::key

qarkai:


caclark:
Implemented in commit 1a1fcb97c1b5f64b0178e84b7e26236c0ae4ce82

PR #2275, opened 2026-04-09
Split PanItem::borders

qarkai:


caclark:
Implemented in commit 146f9fa7c6538ab2475ca085344b5b52e3d077e8

PR #2280, opened 2026-04-15
Rename PanItem::color2 to PanItem::border_color

qarkai:
Use constants in `pan_item_image_draw()`.
Also reorder `pan_item_box_new()` parameters.


caclark:
Implemented in commit a1ef50f37e40d726631d010409e0b572b76c4071

Issue #2281, opened 2026-04-16
Security: Heap-buffer-overflow in image-load-psd.cc


tantelinirinatsioriharimanana-lang:
1. Vulnerability Summary
A heap-based buffer overflow vulnerability exists in the PSD image parser of Geeqie (specifically in image-load-psd.cc). When processing a malicious PSD file using RLE compression, the parser trusts a scanline length value from the file without proper validation.
This leads to an application crash (DoS) and potentially allows for arbitrary code execution.

2. Environment Specifications

    Product: Geeqie 2.7.1
    OS: Ubuntu 22.04.4 LTS
    Compiler: GCC 11.4.0 (with AddressSanitizer)
    Architecture: x86_64

3. Root Cause Analysis
The vulnerability occurs in the PSD_STATE_CHANNEL_DATA state. The code reads line_length directly from the file's lines_lengths array without bounds checking.
The temporary buffer ctx->buffer is allocated based on:
ctx->width * 2 * ctx->depth_bytes
In the tested case:

    Buffer size allocated: 10,000 bytes.
    Malicious line_length from file: 32,579 bytes.
    Result: memcpy writes 32,579 bytes into the 10,000-byte buffer, causing a heap overflow.

4. GDB & ASan Proof of Concept
Context at crash:

(gdb) p *ctx
$1 = {
    state = PSD_STATE_CHANNEL_DATA,
    width = 5000,
    height = 5000,
    compression = PSD_COMPRESSION_RLE,
    ...
}
(gdb) p ctx->width * 2 * ctx->depth_bytes
$3 = 10000  // Allocated buffer size
(gdb) p line_length
$2 = 32579  // Value read from malicious file

AddressSanitizer Report:

==19188==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x526000080810
WRITE of size 32579 at 0x526000080810 thread T8
    #0 0x7ffff743a2c2 in __interceptor_memcpy
    #1 0x5555557ef85b in feed_buffer ../src/image-load-psd.cc:202
    #2 0x5555557f1940 in write ../src/image-load-psd.cc:451

0x526000080810 is located 0 bytes to the right of 10000-byte region [0x52600007e100,0x526000080810)

5. Steps to Reproduce

   1. Create a PSD file with valid headers (e.g., 5000x5000, RGB, 8-bit).
    2.Set the compression method to RLE.
[...]

PR #2285, opened 2026-04-17
Updates image-test.py to always fail when Geeqie crashes, even if we're running an image that's expected to fail.

xsdg:
So, for example, a Geeqie segfault might have return code 134.  Without this change, that would be considered an "expected failure", by virtue of being a non-zero return code.  But with this change, it would still be considered an error, given that it's not 0 (no error) or 1 (Geeqie failed to load the image, but did not crash).

See https://mesonbuild.com/Unit-tests.html#skipped-tests-and-hard-errors .  This is a workaround for a feature introduced in meson 1.11.0

This is intended to allow us to distinguish a Geeqie crash (pre-fix) from an expected-failure (post-fix) for the #2281 test image

### DIFF
--- a/build-aux/image-test.py
+++ b/build-aux/image-test.py
@@ -157,8 +157,22 @@ def main(argv) -> int:
 
 if __name__ == "__main__":
     try:
-        exit(main(sys.argv))
+        exit_code = main(sys.argv)
+        if exit_code not in [0, 1]:
+            # This is so that Geeqie crashes in expected-failure cases will
+            # still trigger a non-passing result.  This is following the
+            # protocol that Meson describes here:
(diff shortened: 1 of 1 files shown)


xsdg:
(From updated commit message)

Without this change:
```
40/59 geeqie:functional+image / Image_ fail_invalid_rle.psd   EXPECTEDFAIL     5.06s   (exit status 134 or signal 6 SIGABRT)
```

With this change:
```
34/59 geeqie:functional+image / Image_ fail_invalid_rle.psd   ERROR            5.07s   exit status 99
```

caclark:
Implemented in commit d61b59a293de193aa42aa91f27254ba7d95014e7

PR #2286, opened 2026-04-18
Use GqPoint in pixbuf_draw_line()

qarkai:


caclark:
Implemented in commit c13601bd15c18327f95be97fb98792bb38d1f74e

PR #2287, opened 2026-04-18
Calculate maximum label width in PanTextAlignment::add()

qarkai:


caclark:
Implemented in commit b4101b770efec290553f04e9e325db8ece3305c9

PR #2289, opened 2026-04-18
Reduce scope of result GdkRectangle for gdk_rectangle_intersect()

qarkai:


caclark:
Implemented in commit e1c6fb81b3b95959b836fdf5fc6e91de815c80b2

Issue #2261, opened 2026-03-25
Thumb maintenance leaks memory [patch attached]


installgentoo:
**Setup (please complete the following information):**
 - Distribution: gentoo
 - Geeqie version [geeqie --version]:Geeqie 2.4 GTK3

**Describe the bug**
Never frees pixbufs loaded for thumb maintenance -> render thumbs. 16gb OOM at 80k files. Rather serious issue.

**Additional context**
Patch attached. Editorialise and commit it.

installgentoo:
[memleak.patch](https://github.com/user-attachments/files/26234993/memleak.patch)

@caclark

PR #2291, opened 2026-04-21
Adds some infrastructure to help debug FileData leaks and refcounting issues

xsdg:
This general approach was suggested by @caclark on the Geeqie mailing list:
https://www.freelists.org/post/geeqie/FileData-leaks

The topic of FileData leaks in Geeqie was raised in #2261

The current version of the test does not auto-detect FileData leaks, but it does still make it significantly easier to notice and debug the leaks that are present.


caclark:
Implemented in commits
02b8394015f0b310e88681c944effb8db9893f99
140a6d08d9e9dd99666a0b285e0631c478f3f3a0
160071a8dc84c06865325038b7f79cb687991870

PR #2292, opened 2026-04-21
Modifies image-test to symlink the image being tested into an otherwise-empty directory

xsdg:
Without this change, the fd_verbose_debug mode showed that Geeqie was creating a FileData for every image in the images directory.  So if creating a FileData for any single image caused Geeqie to crash, that crash might have also shown up in tests for unrelated images.


caclark:
Implemented in commit bb80ea1406b4e751505e181c10d84b1ad1d61f65

PR #2294, opened 2026-04-21
Fix memory leak in pan_queue_image_done_cb()

qarkai:


caclark:
Implemented in commit cc8ef091b77c7748acce1c6b96fab3baf6d02a3a

PR #2300, opened 2026-04-26
Drop unused pan_window_list

qarkai:


caclark:
Implemented in commit 7382a1e625b41cdaf17a0f83d53d9823dcaeba4d

PR #2301, opened 2026-04-26
Use PanViewFilterElement for PanFilterCallbackState::filter_element

qarkai:
Also inline `pan_filter_callback_state_free()`.


caclark:
Implemented in commit c83d91849a9567105b0e8703ffb41c80cbdb7c5e

PR #2305, opened 2026-05-01
Return PanItemList from pan_search_by_date_val()

qarkai:
Simplify related code a bit.


caclark:
Implemented in commit bf0f2b564c458c5f954cd4c67b17b15effe24996

PR #2306, opened 2026-05-01
Use PanItemList for PanWindow list, list_static, queue

qarkai:
Also in related PanGrid::list and FlowerGroup::items.


caclark:
Implemented in commit 2057f29a5842c4a0a518233795b1c93fceecee07

PR #2307, opened 2026-05-01
Inline pan_layout_intersect_l()

qarkai:


caclark:
Implemented in commit e3fc1323e4bdf2cac7ed14f096264c61129122be

PR #2308, opened 2026-05-01
Extract pan_parse_date() from pan_search_by_date()

qarkai:


caclark:
Implemented in commit 36147087179d4705db693e8443d62942b8862dbb

PR #2310, opened 2026-05-02
Use unique_ptr<FlowerGroup> in pan_flower_group()

qarkai:
to avoid manual memory clear.


caclark:
Implemented in commit cbd5e2ac388b27d7c2a047f65f3505759c4cfff2

PR #2311, opened 2026-05-02
Use std::list for FlowerGroup::children

qarkai:


caclark:
Implemented in commit 3e857ed7cb212af67f08b9a72bce6dac0bb89950

PR #2313, opened 2026-05-02
Fix FileData leak in pan_filter_fd_list()

qarkai:


caclark:
Implemented in commit ce7aa329781a0476e0ab1bc177ac796c176a2362

PR #2314, opened 2026-05-03
Move type and key definition to pan_search_activate_cb()

qarkai:


caclark:
Implemented in commit 89ac9b60ca5289acb1b425265d09f42265a87b43

PR #2315, opened 2026-05-03
Add pan_queue_pi_done() to reduce code duplication

qarkai:


caclark:
Implemented in commit 104485716924746db80cca333651ab9107831214

PR #2317, opened 2026-05-04
Use cache_sim_data_set_dimensions() to reduce code duplication

qarkai:


caclark:
Implemented in commit f58bc2b5d534ff51fc0c0f058d844e14df6c06fb

PR #2322, opened 2026-05-09
Use GqSize for CacheData::dimensions

qarkai:
Rename bool `CacheData::dimensions` to `CacheData::have_dimensions`.


caclark:
Implemented in commit 6409e1ba1aed96c5ecd2a5a7cd07469f3cbb680f

PR #2323, opened 2026-05-09
Use GqSize in cache_sim_data_set_dimensions()

qarkai:


caclark:
Implemented in commit 16ffc6dcd96c983674925dc85dafdae99f64f21e

PR #2327, opened 2026-05-10
Split cache_sim_data_new()

qarkai:
Simplify `CacheData::load()` a bit.


caclark:
Implemented in commit 02b600dcb114b2f5bd56d7c98cbd890cdb2e9a21

PR #2328, opened 2026-05-10
Use unique_ptr for CacheData::sim

qarkai:
Also rename `CacheData::similarity` to `CacheData::have_similarity`.


caclark:
Implemented in commit ca424a5331163309c151556060d0312d9f4c2eb6

PR #2329, opened 2026-05-10
Use more std::unique_ptr<ImageSimilarityData>

qarkai:


caclark:
Implemented in commit 92a2b4d250fa8b6f13f4ed97b7d5d2e105612936

PR #2331, opened 2026-05-10
Drop image_sim_new_from_pixbuf()

qarkai:
Convert `image_sim_fill_data()` to `ImageSimilarityData::fill_data()` and simplify a bit.
Also change `CacheData::set_similarity()` parameter type.


caclark:
Implemented in commit a210dc1d68c84ed3ea7755b1e5bf27e3b17d066f

PR #2332, opened 2026-05-10
Drop redundant CacheData::path member

qarkai:


caclark:
Implemented in commit ee78c8cf244f7a5400e5810776f659086100bcd5

PR #2333, opened 2026-05-10
Use less cache_sim_data_free()

qarkai:


caclark:
Implemented in commit 77d240257f3a32fb6972e17f520574cc37a2790a

PR #2334, opened 2026-05-10
Convert CacheType to enum class

qarkai:


caclark:
Implemented in commit b804d44ee25324f04316c9e2c313cb610c32eaac

PR #2336, opened 2026-05-11
Move cache path building to CacheData::load() and CacheData::save()

qarkai:


caclark:
Implemented in commit e78d77a0b394052050cb66301dbe7cdd78983cf8

PR #2339, opened 2026-05-12
Drop redundant CacheData::have_similarity

qarkai:
Rename `CacheData::sim` to `CacheData::similarity`.


caclark:
Implemented in commit ab882cb8fa0037ad2f1a0e670b0e6d2ead11f128

PR #2340, opened 2026-05-12
Fixes a use-after-free bug in FileCache where a pointer was invalidated by the notify callback system

xsdg:
See the analysis in #2335 .  This fixes the use-after-free bug (which caused the crashes), and adds a basic unit test that reliably triggered the bug.  Also fixes an unrelated bug in the same function.

It looks like this bug was introduced in 2012, but went unnoticed for so long because it's predominantly triggered when (1) thumbnail modes are turned on -- which is not the default configuration, and (2) images are rendered (triggering `file_cache_get`) when those FileData point to files that no longer exist on disk.

Fixes #2335
Fixes #1221

PR #2343, opened 2026-05-14
Fixes infinite loop + memory exhaustion when reentrant calls to file_cache_get occur

### LINKED ISSUE #2342: Infinite loop + memory exhaustion when viewed files are deleted
 - Distribution: [e.g. Ubuntu] Debian
 - Distribution release: [e.g. 20.04] unstable
 - Geeqie version [geeqie --version]: git ee78c8cf244f7a5400e5810776f659086100bcd5 + fixes in #2340 

**Describe the bug**
Discovered this when validating the fix for #2335 .  Now that Geeqie no longer segfaults, it stops responding, spins on the CPU, and consumes all available memory instead 🙃

**To reproduce**
Very similar steps as the repro case for #2335 :
1) Create a directory /tmp/gq_test/imgs/ containing the geeqie test images
2) Start geeqie and navigate to that directory
3) Ensure that thumbnailing is enabled (in my case, List view mode with "Show thumbnails" enabled.
4) Run the following command to create a bunch of (symlink) files, wait 1 second, and then delete them all:

    ```
    sleep 1; for i in $(seq 1 20); do for file in {a,I,J,R,s}*.*; do ln -s "$file" "xxAA${i}_${file}"; done; done; sleep 2; rm xx*
    ```
5) While step 4 is running, start scrolling down (towards "xx...") in Geeqie

**Expected behavior**
Geeqie should remain responsive and show the remaining files in the directory.

xsdg:
Depends on prior PR #2340 

Also adds significant documentation around the constraints for `file_cache_get` to operate correctly, as well as a unit test that reproduces and confirms the fix for #2342 

Fixes #2342

### DIFF
--- a/src/filecache.cc
+++ b/src/filecache.cc
@@ -39,6 +39,7 @@ namespace
 struct FileCacheEntry {
 	FileData *fd;
 	gulong size;
+	gboolean checking_if_changed;
 };
 
 #ifdef DEBUG
@@ -126,6 +127,24 @@ FileCacheData *file_cache_new(FileCacheReleaseFunc release, gulong max_size)
 
 gboolean file_cache_get(FileCacheData *fc, FileData *fd)
 {
+	/* Operating theory of this function:
+	 * This function must be re-entrant, which means it must specifically be implemented in a
+	 * way that remains correct even when a re-entrant call happens.
+	 *
+	 * In particular, the file_data_check_changed_files function call may trigger another call
+	 * into this function.  In order to handle that case correctly, we establish that if the
+	 * FileData has changed (in which case we plan to evict it and return FALSE), any
+	 * re-entrant calls into the function will return TRUE _without_ checking for changes.
+	 * Then we will evict the FileData from the cache (if that hasn't happened already), and
+	 * then we will return FALSE.
+	 *
+	 * That said, because it's also possible for a re-entrant call to target a _different_
+	 * FileData than the one that we plan to evict, we stash a checking_if_changed bool in
+	 * every FileCacheEntry.  That is, we need to handle the case where
+	 * file_cache_get(fc, fd_A) triggers file_cache_get(fc, fd_B), which in turn triggers
+	 * file_cache_get(fc, fd_A) again.
+	 */
+
 	g_assert(fc && fd);
 
 	GList *work = g_list_find_custom(fc->list, fd, reinterpret_cast<GCompareFunc>(file_cache_entry_compare_fd));
@@ -135,10 +154,10 @@ gboolean file_cache_get(FileCacheData *fc, FileData *fd)
 		return FALSE;
 		}
 
-	/* entry exists */
+	// Entry exists.
 	DEBUG_2("cache hit: fc=%p %s", (void *)fc, fd->path);
 
-	/* move it to the beginning, if needed */
+	// Move it to the beginning, if needed.
 	if (work != fc->list)
 		{
 		DEBUG_2("cache move to front: fc=%p %s", (void *)fc, fd->path);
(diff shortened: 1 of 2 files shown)


xsdg:
Finally 🥵

The repro steps from #2342 _do_ cause Geeqie to stall for a bit, which is probably worth digging into, but it eventually comes out of it and goes back to being functional.  It no longer crashes or ends up in an infinite loop.

src/filecache.cc
```diff
@@ -39,6 +39,7 @@ namespace
 struct FileCacheEntry {
 	FileData *fd;
 	gulong size;
+	gboolean checking_if_changed;
```
qarkai, line 42: Suggestion: prefer `bool` instead of `gboolean` unless variable is used in GTK functions.

`Use glib types when possible (ie. gint and gchar instead of int and char).` from https://github.com/BestImageViewer/geeqie/blob/master/CODING.md#source-code-style is obsolete. Documentation recommends using standard C types, see e.g. https://docs.gtk.org/glib/types.html#gint.

caclark:
> Suggestion: prefer bool instead of gboolean unless variable is used in GTK functions.

If there is a consensus the Geeqie Coding Standards should be updated.

src/filecache.cc
xsdg: Understood.  I feel like my preference would be for this stack of commits to be merged as-is for right now.  I'm planning on doing a refactor to switch this file from GList to std::vector (to improve type safety), and that would be a reasonable time to also switch other types (including this one) to their C/C++ counterparts.

Tagging @caclark since I don't know how notifications on github work 🙃

caclark:
@xsdg 
Unfortunately there are conflicts here.

xsdg:
Yeah, rebased.  Unfortunately, GitHub's stacked PR workflow is still in closed beta (https://github.github.com/gh-stack/), so there's no good way to avoid this at the moment.

xsdg:
Regarding the failing checks, it looks like something's up with heif/heic support.  This passes on my local machine (and nothing in this PR should affect only individual formats), so I'm not sure what's up.

caclark:
@xsdg 

> This passes on my local machine

Using the latest .diff and running `./tools/test-all.sh` I get multiple errors :
 23/1545 functional+image - geeqie:Image_ RAW_PANASONIC_DMCL10.RAW                      FAIL             5.89s   exit status 1
  30/1545 functional+image - geeqie:Image_ Sample.ani                                    FAIL             7.80s   exit status 1
  31/1545 functional+image - geeqie:Image_ Sample.qtif                                   FAIL             7.79s   exit status 1
  34/1545 functional+image - geeqie:Image_ autumn_1440x96000.heic                        FAIL             7.71s   exit status 1
  35/1545 functional+image - geeqie:Image_ cheers_1440x960.heif                          FAIL             7.72s   exit status 1
  48/1545 functional+image - geeqie:Image_ sample2.pdf                                   FAIL             8.80s   exit status 1

How do you run tests locally?

xsdg:
I was not aware of `src/tools/test-all.sh`, but all checks are currently passing on github (see <https://github.com/BestImageViewer/geeqie/actions/runs/25964962716?pr=2343>).

My comment above was re #2347, which was a github-specific failure.  Locally, I don't have all of the libraries installed, so when I compile/test geeqie locally, the image tests consistently fail only for libraries that aren't installed.  But based on the github passing, I would expect local tests to also pass if I had those libraries installed.

Locally, I see this result, which is as expected on my machine:
```
Summary of Failures:

   4/1421 geeqie:functional+image / Image_ IMG_20200308_194050.jxl                       FAIL             6.28s   exit status 1
   8/1421 geeqie:functional+image / Image_ M_92_Light_R_001.fits                         FAIL             6.34s   exit status 1
  13/1421 geeqie:functional+image / Image_ GammaChart.exr                                FAIL             8.29s   exit status 1
  45/1421 geeqie:functional+image / Image_ sample1.jp2                                   FAIL             6.25s   exit status 1
  47/1421 geeqie:functional+image / Image_ sample2.djvu                                  FAIL             6.26s   exit status 1
  48/1421 geeqie:functional+image / Image_ sample2.pdf                                   FAIL             6.29s   exit status 1
 131/1421 geeqie:analysis / Code Correctness_ similar.cc                                 FAIL             9.06s   exit status 1
1419/1421 geeqie:analysis / Ancillary files                                              FAIL             1.03s   exit status 1
```

The image tests fail because of libraries I don't have installed.  `similar.cc` fails because of a compiler change that isn't in our test compilers yet.  `Ancillary files` fails because I don't have `shellcheck` and the other static analyzers installed.

caclark:
Implemented in commit c126e4f4c4a5081dc84de1dd77ec63546e9b81d9 f82ad6ee0624cdf9ec5ae0b8e19cb37ede24fdf1

PR #2345, opened 2026-05-14
Remove obsolete coding rule

qarkai:


qarkai:
@caclark @xsdg If you agree with bool over gboolean suggestion I'll add it in this PR.

caclark:
Implemented in commit a536fd4c9ebabd340bbf2613d9c9a0a97a7d2177

PR #2346, opened 2026-05-14
Remove comparison with gboolean literals

qarkai:
See https://docs.gtk.org/glib/types.html#gboolean for rationale.


caclark:
Implemented in commit f256839e7da7b0fc5a14c4e9e85c64b21bebc4c6

PR #2352, opened 2026-05-18
Simplify get_exec_time() a bit

qarkai:


caclark:
Implemented in commit 424cc9ae88aa8098cd610b8eee53276e0cffe068

PR #2353, opened 2026-05-18
Replace deprecated gettimeofday() with std::chrono

qarkai:


caclark:
Implemented in commit 997e5c87399ee52b762daa8b814066a36fb9c1b0

PR #2354, opened 2026-05-19
Use unique_ptr for PanCacheData::cd

qarkai:


qarkai:
Weird warning
```
../src/pan-view/pan-view.cc:602:16: error: unchecked access to optional value [bugprone-unchecked-optional-access,-warnings-as-errors]
  602 |                                 fd->date = pc->cd->date.value();
      |                                            ^
```
`pc->cd->date` is checked right before. Let's try `.has_value()`.

qarkai:
It is not about `.has_value()`. Let's check `pc`.

qarkai:
I'm out of ideas.
@caclark @xsdg any thoughts how to fix this except NOLINT?

caclark:
@qarkai 
Not from me, but from ChatGPT.....


clang-tidy is not proving that this check:

pc->cd->date && pc->cd->date >= 0

guarantees safe .value() access.

Use value_or() or store the optional first.

Best simple fix:

if (needle)
	{
	auto *pc = static_cast<PanCacheData *>(needle->data);

	if (pc && pc->cd)
		{
		const auto date = pc->cd->date.value_or(-1);

		if (date >= 0)
			{
			fd->date = date;
			}
		}

	haystack = g_list_delete_link(haystack, needle);
	}

Or, if you prefer keeping the explicit optional check:

if (needle)
	{
	auto *pc = static_cast<PanCacheData *>(needle->data);

	if (pc && pc->cd)
		{
		const auto &date = pc->cd->date;

		if (date.has_value() && date.value() >= 0)
			{
			fd->date = date.value();
			}
		}

	haystack = g_list_delete_link(haystack, needle);
	}

But value_or(-1) is cleaner here because your logic already treats negative dates as invalid.

qarkai:
`.value_or()` trick works. Thank you, @caclark!

caclark:
Implemented in commit f32b720142b6d42d906550f83d29eb1fe76a7939

PR #2358, opened 2026-05-23
Use unique_ptr for DupeItem::simd

qarkai:
Remove now unused `image_sim_free()`.


caclark:
Implemented in commit 030d418795afb733aed0954e1e94ff3f6bcfba24

PR #2359, opened 2026-05-23
Replace image_sim_new() with constructors

qarkai:


caclark:
Implemented in commit ed3e79820507514880ea9d0ce49a72b8a5c8e497

PR #2361, opened 2026-05-23
Use bool for ImageSimilarityData::filled and image_sim_filled()

qarkai:
Use more `image_sim_filled()` instead of manual `ImageSimilarityData::filled` check.


caclark:
Implemented in commit 061ceb944d5535de20bb5fa73277a029ba9317e0

PR #2362, opened 2026-05-23
Splits geeqie execution from test-ancillary-files into its own test, and runs that test in an isolated environment

xsdg:
This change maintains a (relatively) hermetic environment for `test-bash-completion`, which avoids interactions with or dependencies on any other geeqie installations on the test system.

Note that this PR does some fancy merge tricks to maintain per-line blame history in both `test-ancillary-files.sh` as well as `test-bash-completion.sh`.  It cannot be rebased.  For more details, see:
<https://devblogs.microsoft.com/oldnewthing/20190917-00/?p=102894>


caclark:
@xsdg  



Sorry... I cannot apply this patch:

cclark@office:~/data/geeqie/geeqie$ git am ~/Downloads/2362.patch
Applying: Removes the contents that are being split out into test-bash-completions.sh
Applying: Creates test-bash-completions.sh from test-ancillary-files.sh
Applying: Maintains the relevant parts of test-bash-completions.sh
error: patch failed: build-aux/test-bash-completions.sh:21
error: build-aux/test-bash-completions.sh: patch does not apply
Patch failed at 0003 Maintains the relevant parts of test-bash-completions.sh
hint: Use 'git am --show-current-patch=diff' to see the failed patch
hint: When you have resolved this problem, run "git am --continue".
hint: If you prefer to skip this patch, run "git am --skip" instead.
hint: To restore the original branch and stop patching, run "git am --abort".
hint: Disable this message with "git config set advice.mergeConflict false"

xsdg:
@caclark I can try doing this directly, if you'd prefer.  It looks like I still have access to the upstream repository.

That said, GitHub shows some instructions that should work here.  if you look at the bottom of this page, it says "This branch cannot be rebased due to conflicts," and below that is a link, "You can also merge this with the command line.  **View command line instructions.**"  Those instructions should work for you, including from a clone of the upstream repository

```
git checkout -b xsdg-split_comp_test master
git pull https://github.com/xsdg/gqpp.git split_comp_test

git checkout master
git merge --no-ff xsdg-split_comp_test
git push <remote> master
```

Unfortunately, the fancy commit structure for this trick can't be represented by a patch, which is why using a merge is necessary.

caclark:
@xsdg 

> I can try doing this directly, if you'd prefer.

OK. Better than me destroying things unwittingly.

PR #2363, opened 2026-05-24
Return string from md5_digest_to_text()

qarkai:


caclark:
Implemented in commit 2739826fbb61e87742dc67f8a22846529229b2d9

qarkai:
Sorry, I messed up with string initialization using curly braces. It should be parentheses. I'll fix it in the next PR.

PR #2364, opened 2026-05-25
Fix string initialization

qarkai:


qarkai:
I wonder why use a binary digest at all? Just to reduce the size of `CacheData`? Difference is 16 bytes.

caclark:
Implemented in commit 4a7dce0456c5966459aac52911d72021e92bc292

PR #2369, opened 2026-05-26
Use optional string for DupeItem::md5sum

qarkai:
Return `std::string` from `md5_get_string_from_file()`.


caclark:
Implemented in commit e799b1f38e725d6088a3c2afb9647b9b424cbc68

PR #2370, opened 2026-05-27
Switches FileCache implementation from C/glib style to C++ style

xsdg:
Among other things, this allows instantiating FileCache on the stack instead of only on the heap, and allows it to be constructed and destructed using C++-style idioms.  This also fully transitions from glib types to C++ types.

### DIFF
--- a/src/exif-common.cc
+++ b/src/exif-common.cc
@@ -853,7 +853,7 @@ ExifData *exif_read_fd(FileData *fd)
 {
 	if (!fd) return nullptr;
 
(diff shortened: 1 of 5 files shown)


caclark:
Implemented in commit 68d9dacfb5b787cee3cd6aa2a542455fbd27c0bb c5f643976948dd84bfb356b657a6993d17e46385 05b62dfe6009671f777f5ab5473cc82b241238e6 71d135fb81497131892c834525ded02e562af5b5 8ab2d2f9802968c270b8f984745182e09037c4f0

PR #2374, opened 2026-05-27
Use unique_ptr for CacheLoader::cd

qarkai:


caclark:
Implemented in commit 2a2f5b0cc3e69d7347dd1d2d5a801c2a27c869e3

PR #2379, opened 2026-05-30
Use unique_ptr for SearchData search_similarity_cd and img_cd

qarkai:


qarkai:
`SearchData` uses `ImageLoader` to load `CacheData`. Is there reason for not using `CacheLoader`?

@caclark @xsdg @installgentoo any thoughts?

caclark:
Implemented in commit 481831736085a77d8d876c8128e4412f35732cbe

PR #2381, opened 2026-06-01
Simplify cache_loader_phase1_process() a bit

qarkai:
Do not try to load image if similarity is not required.


caclark:
Implemented in commit 4bf957470d46a717a8e61912ffa8a5a951fd82df

PR #2383, opened 2026-06-04
Fix memory leak in process_files()

qarkai:
Make parameters const in related functions.


caclark:
Implemented in commit 7c86bc65071a110d9b18bd6a4b3d6715f98308c7

PR #2384, opened 2026-06-06
Use GqSize for DupeItem::dimensions

qarkai:
Rename old `DupeItem::dimensions` to `DupeItem::dimensions_sum`.


caclark:
Implemented in commit 4589621039f0977011288cfaa71f974472dc0c28

PR #2385, opened 2026-06-06
Use GqSize for SearchData search dimensions

qarkai:
Simplify `SearchUi` dimensions.


caclark:
Implemented in commit 49aab7c841b03ac3617147fe8bcd9995d6ca54c3

PR #2386, opened 2026-06-06
Fix memory leak in column_sort_cb()

qarkai:
Simplify a bit.


caclark:
Implemented in commit 6eaf5545d4df250659fd9e97f7dba77ecd563d18

Issue #2293, opened 2026-04-21
Geeqie FileData leaks identified by refcount debugging mode


xsdg:
**Setup (please complete the following information):**
 - Distribution: [e.g. Ubuntu] Debian
 - Distribution release: [e.g. 20.04] Sid
 - Geeqie version [geeqie --version]: git a1ef50f3

**Describe the bug**
FileData leaks with the following refcount sequences:
```
FileData dump start
/path/to/.../build/src/../../src/main.cc:528 exit_program_final

2
2
2    /tmp/tmp.QUXpoOpmo4/container_dir/images_dir/DSCF0253.MPO
    r -> 2 @ ../src/view-file/view-file-list.cc:845
    r -> 3 @ ../src/image.cc:753
    r -> 4 @ ../src/image-load.cc:298
    r -> 5 @ ../src/filecache.cc:168
    r -> 6 @ ../src/filedata/filedata.cc:372
    u -> 5 @ ../src/filedata/filelist.cc:356
    r -> 6 @ ../src/filecache.cc:168
    u -> 5 @ ../src/image-load.cc:275
    r -> 6 @ ../src/filedata/filedata.cc:372
    u -> 5 @ ../src/filedata/filelist.cc:356
    r -> 6 @ ../src/filedata/filedata.cc:372
    u -> 5 @ ../src/filedata/filelist.cc:356
    u -> 4 @ ../src/view-file/view-file-list.cc:154
    u -> 3 @ ../src/filedata/filelist.cc:356
    u -> 2 @ ../src/image.cc:690
3    /tmp/tmp.QUXpoOpmo4/container_dir/images_dir/Chimera-AV1-8bit-768x432-1160kbps-162.avif
    r -> 2 @ ../src/view-file/view-file-list.cc:845
    r -> 3 @ ../src/image.cc:1101
    r -> 4 @ ../src/image-load.cc:298
    r -> 5 @ ../src/filecache.cc:168
    r -> 6 @ ../src/filedata/filedata.cc:2800
    r -> 7 @ ../src/filedata/filedata.cc:372
    r -> 8 @ ../src/view-file/view-file-list.cc:1238
    u -> 7 @ ../src/filedata/filelist.cc:356
    u -> 6 @ ../src/filedata/filelist.cc:356
    u -> 5 @ ../src/image-load.cc:275
    r -> 6 @ ../src/image-load.cc:298
    r -> 7 @ ../src/filecache.cc:168
[...]

PR #2391, opened 2026-06-08
Extends `FileDataRef` to be a managed ref-counting shared pointer for `FileData`, and the default storage for `FileData` instances.

xsdg:
This makes it easier to hold a `FileData *` (whether persistently in an object or within a given scope) while avoiding `FileData` leaks.

Note that this will only `unref` correctly when created on the stack, or when the containing object uses `new/delete` semantics -- the destructor will **not** be called by any variant of `g_free`.

This is in response to discussion in #2293

### DIFF
--- a/src/filedata.h
+++ b/src/filedata.h
@@ -188,6 +188,42 @@ struct FileDataDebugInfo {
 };
 #endif
 
+class FileData;
+
+/**
+ * @class FileDataRef
+ * @brief RAII object that holds a ref to the specified FileData.
+ */
+class FileDataRef
+{
+    public:
+	explicit FileDataRef(FileData *fd);
+	// Not copyable or copy-assignable.
+	FileDataRef(FileDataRef &) = delete;
+	FileDataRef &operator=(const FileDataRef &) = delete;
+
+	// Allow move and move-assignment.
+	FileDataRef(FileDataRef &&) noexcept;
+	FileDataRef &operator=(FileDataRef &&) noexcept;
+
+	~FileDataRef();
+
+	// Allow implicit conversion directly to FileData*.
+	// This allows a FileDataRef to be used anywhere a FileData* is expected.
+	operator FileData*() const { return fd_; }
+
+	// In a boolean context, return whether or not we're holding nullptr.
+	operator bool() const { return fd_ != nullptr; }
+
+	void reset(FileData *new_fd);
+	FileData *release();
+	FileData* operator*() const { return fd_; }
+	FileData* operator->() const { return fd_; }
+
+    private:
+	FileData *fd_ = nullptr;
+};
+
 class FileData {
 private:
 	FileData() = delete;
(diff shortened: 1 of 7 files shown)


xsdg:
@qarkai does this do what you need?

src/filedata.h
```diff
@@ -482,19 +482,27 @@ class FileData::FileList
 };
 
 /**
- * @class FileDataRef
+ * @class FileDataRefPtr
  * @brief RAII object that holds a ref to the specified FileData.
```
qarkai: Doesn't hold ref anymore.

`FileDataRefPtr` is cumbersome. Now it holds pointer, so why not just `FileDataPtr`?

src/filedata/filedata.cc
```diff
@@ -3010,14 +3010,24 @@ bool FileData::supports_exif_orientation() const
...
 }
 
-FileDataRef::~FileDataRef()
+FileDataRefPtr::~FileDataRefPtr()
 {
-        fd_.file_data_unref();
+	if(fd_ != nullptr) fd_->file_data_unref();
+}
+
+void FileDataRefPtr::reset(FileData *new_fd)
```
qarkai: What if `new_fd` is result of `file_data_new_something()`? I mean

```cpp
fd_ref.reset(file_data_new_simple(path));
```

src/filedata.h
xsdg: Yeah, there's some really unfortunate ambiguity here (very open to suggestions).  My intent in the original naming was that it holds a (`FileData` refcounting) ref.  It was only a coincidence that it happened to hold a C++ reference to the `FileData` object.

And so the intent of this name is that it is a pointer-like object that still holds a `FileData` refcounting ref.  But it makes sense that this would be very confusing, in the context of C++ references and pointers.

I don't think it's intuitive from the name that a `FileDataPtr` object would have refcounting side-effects, which is why I didn't just go with that name.

Just brainstorming: `FileDataPtrWithRef`?  `FileDataRefCountPtr`?  I don't feel like these are good names, but I'm also not sure what a better direction would be.

src/filedata/filedata.cc
xsdg: I feel like the options are to just have two different `reset` methods, or to add a bool arg to the single method (less good imo).  My brain is currently bouncing around `reset` vs `reset_without_ref` vs `reset_with_ref` vs `reset_and_ref`.  Thoughts?  Other suggestions?

qarkai: It seems that issue in trying to implement two usecases in one class. On one hand current `FileDataRef` behavior: ref on construction (I know about `skip_ref` flag), unref on destruction. `FileDataRef` adds and holds own reference (in FileData terms, not C++) of given `FileData`. This reminds shared_ptr when passing it as copy.
On other hand - desired behavior: taking ownership on construction (`FileDataRef` with `skip_ref=TRUE`), unref on destruction. No added reference, just hold and "destroy" what was given. This is exactly what unique_ptr does.
I guess in most cases we need unique_ptr-like behavior. I suggest to split new behavior to separate class. It could be hand-written or simple alias to unique_ptr with custom deleter. I don't have good idea for name, could be `FileDataOwner`, `FileDataPtr` or something.
Important: `skip_ref` flag should be removed from `FileDataRef`, always ref given `FileData` in constructor and `FileDataRef::reset()`.

xsdg: (for simplicity, I'm going to reply to this part of the conversation together with your points in #2293.  Leaving this comment open so it's clear that this is unresolved)

xsdg: Okay, updated.  I still want to reorganize the unit tests a bit and make the `FileData::new` function names more consistent, but this feels like it should handle both usecases with an API that is pretty clear and easy to use.  I'm curious what you think about this version.  (And also, will squash some commits before taking the PR back out of draft mode 👍)

src/filedata/filedata.cc
```diff
@@ -3010,14 +2994,44 @@ bool FileData::supports_exif_orientation() const
...
+	fd_ = new_fd;
+
+	if (old_fd != nullptr) old_fd->file_data_unref();
+}
+
+FileData *FileDataRef::release()
+{
+	FileData *old_fd = fd_;
+	fd_ = nullptr;
+	return old_fd;
```
qarkai:
```suggestion
	return g_steal_pointer(&fd_);
```

qarkai (commented):
LGTM.

Could you make another PR after merging this one where `FileData *` member is replaced with `FileDataRef` in any struct? E.g. `PanCacheData`, it already uses `new`/`delete`.

src/filedata.h
xsdg: (Renamed back to `FileDataRef` as discussed... somewhere... 😅)

src/filedata/filedata.cc
xsdg: (Applied manually; thanks for the tip!)

caclark:
Implemented in commits f60d4db03be931564e551a92e1573e33911b79e6 85158055f8060c841682279a63c9c9752ac2e8be 0ea4b7f7d7bcc692710f7da0c92c6d658faa3a01

qarkai:
@caclark looks like #2404, #2405, #2406, #2407 were closed by mistake.

caclark:
@qarkai  @xsdg 

Sorry, I have no idea what happened.

Do you have a suggestion to revert?

Will a git revert bring back the pull requests?

xsdg:
Oh, this was actually my mistake.

I had accidentally pushed those branches to the main Geeqie github repository instead of my personal fork, so the branches were auto-deleted when Klaus' infrastructure synced the upstream repository here.  I'll just open new pull requests from my fork and everything will be okay.  Sorry about that

PR #2392, opened 2026-06-08
GTK4: Migrate menus and actions\nFor information only

caclark:
This is an additional backup in case my main and backup disks fail. There is too much work to risk being lost.

This is a functional patch, but it is still incomplete.
It is based on commit ef17881f2db94f9dc84f8

This converts the menus and actions from GTK2/GTK3 style to GTK4 style, but still runs under GTK3.

This change is an absolute pain in the neck. I did not find any way to make incremental changes, so this has ended up as an all-or-nothing change. Reviewing the changes is impossible...

If anyone tries this patch, note the warning message regarding config. file changes which is shown on first run.


xsdg:
@caclark You should create your own fork of Geeqie (Look for the "Fork" button in the top-right corner of https://github.com/BestImageViewer/geeqie ).  Then if you clone that fork, you can push things there and use it as a mirror of all of your local changes, without affecting the Github main geeqie repository (and having those changes steamrolled by Klaus' infrastructure)

Then if you want a second pair of eyes on a change, it's easy to point someone else at a branch or pull request that you make from your fork to the main Geeqie github repository (the same way @qarkai and I do)

PR #2393, opened 2026-06-08
Fix FileData leak in process_files()

qarkai:


caclark:
Implemented in commit 5e5850bc629edae6fb90d281d63d2c6b5e1a1947

PR #2394, opened 2026-06-08
Fix return value in download_web_file()

qarkai:


caclark:
Implemented in commit 30b729babdbc179ab1fea294e00be09ce8f3d01e

PR #2397, opened 2026-06-11
Simplify history_list_save() a bit

qarkai:
Reduce number of string comparisons.


caclark:
Implemented in commit 8491aab785f96cf595c7c93d7000252e6209ee95

PR #2398, opened 2026-06-11
Use string for HistoryData::key

qarkai:


caclark:
Implemented in commit 8fbe5d78e6a64260829728a26430cc0e1f307a4c

PR #2400, opened 2026-06-12
Simplify pref_list_get() and related functions a bit

qarkai:


caclark:
Implemented in commit 39f15fc22f83afa7497322333f5f399835d7d939

PR #2402, opened 2026-06-12
Use map for history_list

qarkai:
@xsdg @installgentoo could you review this?

### DIFF
--- a/src/history-list.cc
+++ b/src/history-list.cc
@@ -20,8 +20,10 @@
 
 #include "history-list.h"
 
+#include <algorithm>
 #include <cstdio>
 #include <cstring>
+#include <unordered_map>
 #include <vector>
 
 #include "options.h"
@@ -91,10 +93,12 @@ bool HistoryChain::push_back(const gchar *path)
 HistoryChain history_chain{};
 HistoryChain image_chain{};
 
-gint dirname_compare(gconstpointer data, gconstpointer user_data)
+std::unordered_map<std::string, HistoryList> history_list_map;
+
+bool dirname_compare(const std::string &item, const gchar *path)
 {
-	g_autofree gchar *dirname = g_path_get_dirname(static_cast<const gchar *>(data));
-	return g_strcmp0(dirname, static_cast<const gchar *>(user_data));
+	g_autofree gchar *dirname = g_path_get_dirname(item.c_str());
+	return g_strcmp0(dirname, path) == 0;
 }
 
 } // namespace
@@ -181,15 +185,6 @@ void image_chain_append_end(const gchar *path)
  *-----------------------------------------------------------------------------
  */
 
-struct HistoryData
-{
-	std::string key;
-	GList *list;
-};
-
-static GList *history_list = nullptr;
-
-
 static gchar *quoted_from_text(const gchar *text)
 {
 	if (text[0] == '\0') return nullptr;
@@ -254,34 +249,36 @@ gboolean history_list_save(const gchar *path)
 
 	g_autoptr(GString) gstring = g_string_new("#History lists\n\n");
 
-	for (GList *list = g_list_last(history_list); list; list = list->prev)
+	for (const auto &[key, items] : history_list_map)
 		{
-		const auto *hd = static_cast<HistoryData *>(list->data);
+		g_string_append_printf(gstring, "[%s]\n", key.c_str());
 
-		g_string_append_printf(gstring, "[%s]\n", hd->key.c_str());
-
-		const bool is_recent = (hd->key == "recent");
+		const bool is_recent = (key == "recent");
 
 		/* save them inverted (oldest to newest)
 		 * so that when reading they are added correctly
 		 */
-		GList *work = nullptr;
-		if (hd->key == "path_list")
+		auto last = items.crend();
+		if (key == "path_list")
 			{
-			work = g_list_nth(hd->list, options->open_recent_list_maxsize - 1);
+			if (static_cast<size_t>(options->open_recent_list_maxsize) < items.size())
+				{
+				last = std::next(items.crbegin(), options->open_recent_list_maxsize);
+				}
 			}
-		else if (hd->key == "image_list")
+		else if (key == "image_list")
 			{
-			work = g_list_nth(hd->list, options->recent_folder_image_list_maxsize - 1);
+			if (static_cast<size_t>(options->recent_folder_image_list_maxsize) < items.size())
+				{
+				last = std::next(items.crbegin(), options->recent_folder_image_list_maxsize);
+				}
 			}
 
-		if (!work) work = g_list_last(hd->list);
-		for (; work; work = work->prev)
+		for (auto work = items.crbegin(); work != last; ++work)
 			{
-			const auto *item = static_cast<gchar *>(work->data);
-			if (is_recent && !isfile(item)) continue;
+			if (is_recent && !isfile(work->c_str())) continue;
 
-			g_string_append_printf(gstring, "\"%s\"\n", item);
+			g_string_append_printf(gstring, "\"%s\"\n", work->c_str());
 			}
 		g_string_append(gstring, "\n");
 		}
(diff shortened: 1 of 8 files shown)


src/history-list.cc
```diff
@@ -254,34 +249,36 @@ gboolean history_list_save(const gchar *path)
 
 	g_autoptr(GString) gstring = g_string_new("#History lists\n\n");
 
-	for (GList *list = g_list_last(history_list); list; list = list->prev)
+	for (const auto &[key, list] : history_list)
```
xsdg: Since this is using `std::list` now, maybe rename `list` to something else just to avoid confusion?

caclark:
I use Geany as an editor (plus whatever plugins - it maybe uses Scintilla). There are occasions where the editor highlights a variable as a reserved/compiler name.

Should there be an explicit list of names that should not be used as variable names?

src/history-list.cc
```diff
@@ -91,10 +93,12 @@ bool HistoryChain::push_back(const gchar *path)
 HistoryChain history_chain{};
 HistoryChain image_chain{};
 
-gint dirname_compare(gconstpointer data, gconstpointer user_data)
+std::unordered_map<std::string, HistoryList> history_list;
```
xsdg: I might call this `history_list_map` or `cached_history_lists` or `history_list_cache` or something?  I remember being confused farther down the file when the type retrieved from `history_list` was a `HistoryList`

src/history-list.cc
```diff
@@ -291,161 +288,115 @@ gboolean history_list_save(const gchar *path)
...
-	struct Paths
-	{
-		const gchar *oldpath;
-		const gchar *newpath;
-	} paths{ oldpath, newpath };
+	HistoryList &list = it->second;
+	if (list.empty()) return;
 
-	static const auto find_path = [](gconstpointer data, gconstpointer user_data)
+	const auto find_path = [oldpath, newpath](const std::string &buf)
```
xsdg, line 357: Could you add a brief comment for what this does?  (The code here is easy enough to read, but it's not clear what oldpath and newpath are supposed to represent, which makes it hard to understand the intent of this search function)

xsdg: Actually, if you felt like adding a documentation block for the function, I think that would be pretty helpful for this function in particular.  Up to you for that.

src/history-list.cc
```diff
@@ -291,161 +288,115 @@ gboolean history_list_save(const gchar *path)
...
-	gint p = g_list_position(hd->list, work) + direction;
-	if (p < 0) return;
+	const auto work = std::find(list.cbegin(), list.cend(), path);
+	if (work == list.cend()) return;
 
-	auto buf = static_cast<gchar *>(work->data);
-	hd->list = g_list_remove(hd->list, buf);
-	hd->list = g_list_insert(hd->list, buf, p);
+	if ((direction < 0 && std::distance(list.cbegin(), work) < abs(direction)) ||
+	    (direction > 0 && std::distance(work, list.cend()) < direction))
```
xsdg: I presume this was an intentional behavioral change, to make the error-handling logic symmetric.  Might be something to mention briefly in the commit message, in case other parts of Geeqie were somehow relying on the incorrect original behavior.

src/history-list.cc
```diff
@@ -454,16 +405,15 @@ void history_list_item_remove(const gchar *key, const gchar *path)
 }
 
 /**
- * @brief The returned GList is internal, don't free it
+ * @brief The returned list is internal
```
xsdg: The original comment wasn't useful, other than the "don't free" thing.  At this point, I'd either update the function doc to include more info, or delete it entirely.

src/history-list.cc
```diff
@@ -454,16 +405,15 @@ void history_list_item_remove(const gchar *key, const gchar *path)
...
  */
-GList *history_list_get_by_key(const gchar *key)
+HistoryList &history_list_get_by_key(const gchar *key)
```
xsdg: After reading through https://stackoverflow.com/questions/26858034/why-does-stdoptional-not-have-a-specialization-for-reference-types, would a return type like `std::optional<std::reference_wrapper<HistoryList>>` work?  (I don't actually know)

It seems like the hesitation was around the direct-assignment case, which we can just avoid doing.

xsdg (commented):
I'm not familiar with `HistoryList` functionality in Geeqie, so this review is about the code changes, but doesn't really cover whether this might affect how things work.

Other than the error-handling case I pointed out, this all looks reasonable and doesn't seem to me like it should cause any changes in behavior.

src/history-list.cc
qarkai: Renamed to `items`.

qarkai:
> I use Geany as an editor (plus whatever plugins - it maybe uses Scintilla). There are occasions where the editor highlights a variable as a reserved/compiler name.
> 
> Should there be an explicit list of names that should not be used as variable names?

You mean something like this
https://cppreference.com/cpp/keyword
https://cppreference.com/cpp/language/identifiers#:~:text=with%20the%20following%20exceptions
?

src/history-list.cc
qarkai: Not my logic, just simplified reading.
If you wonder why `.something` is removed from history list, explanation is in 9ccfac429a9bd1a745efd8bc94b6081a7dd6ee23.

Sure, I'll add description.

src/history-list.cc
qarkai: Original behavior was correct too. Container have changed from GList to std::list, so checks have to be adopted.

src/history-list.cc
qarkai: This returns reference to avoid checking result, like in original code. I don't like this `empty_list` too.
Since checking result is ok, I'm going to change this to simple pointer.

src/history-list.cc
qarkai: Or am I missing something about original behavior?

src/history-list.cc
qarkai: `history_list_map` seems nice, thanks.

caclark:
@qarkai 

> You mean something like this

Sorry. I got confused....
It looks like the highlighting is coming from files that give editor highlighting for GTK stuff.

Please ignore...

qarkai:
In general, it's weird that shortcuts, bookmarks, sort_manager, pan_view_options are stored in `history`. They have internal formats, so ini file or key file seems more suitable.

It might make more sense to merge pan_view_options into options.

src/history-list.cc
xsdg: `items` is a really good name for this.  Much easier to follow the logic now

src/history-list.cc
xsdg: From https://docs.gtk.org/glib/type_func.List.insert.html
> position
>   Type: gint
>   The position to insert the element. If this is negative, or is larger than the number of elements in the list, the new element is added on to the end of the list.

So, the behavior is consistent when `direction < 0`: if, for example, `direction < 0 && std::abs(direction) > items->size()`, the move attempt will be rejected with both the original code and the new code.

The behavior change is that if `direction > 0 && direction > items->size()`, with the old code, the target `item` would be moved to the end of the list, whereas now, the move attempt is rejected.  To be clear, I think your current behavior is better and more consistent than the old behavior.  But it's still a change in behavior, if I'm reading things correctly.

src/history-list.cc
```diff
@@ -454,16 +400,14 @@ void history_list_item_remove(const gchar *key, const gchar *path)
...
 {
-	HistoryData *hd;
-
-	hd = history_list_find_by_key(key);
-	if (!hd) return nullptr;
+	if (!key) return nullptr;
 
-	return hd->list;
+	auto work = history_list_map.find(key);
+	return (work != history_list_map.end() && !work->second.empty()) ? &work->second : nullptr;
```
xsdg: No need for the ternary operator here.  This will probably be easier to follow as two lines.
```c++
if ((work != history_list_map.end() && !work->second.empty()) return &work->second;
return nullptr;
```

xsdg (approved):

src/history-list.cc
qarkai: Got rid of ternary operator. Inverted condition though.

qarkai:
@caclark This PR is ready for merge.

caclark:
Implemented in commit be5dac66a22327eb9a58dbd1e227ed636c049a90

PR #2403, opened 2026-06-14
Fixes recent check-build errors

xsdg:
```
./src/bar-keywords.cc:1331:2: error: nested redundant #if; consider removing it [readability-redundant-preprocessor,-warnings-as-errors]
 1331 | #if HAVE_GTK4
      |  ^
```

And

```
../src/advanced-exif.cc:356:8: error: use auto when initializing with a cast to avoid duplicating the type name [modernize-use-auto,-warnings-as-errors]
  356 |         const GdkModifierType event_state = static_cast<GdkModifierType>(event->state);
      |               ^~~~~~~~~~~~~~~
      |               auto
```

### DIFF
--- a/src/advanced-exif.cc
+++ b/src/advanced-exif.cc
@@ -353,7 +353,7 @@ static gboolean advanced_exif_keypress(GtkWidget *, GdkEventKey *event, gpointer
 	const GdkModifierType event_state = state;
 #else
 	const guint event_keyval = event->keyval;
-	const GdkModifierType event_state = static_cast<GdkModifierType>(event->state);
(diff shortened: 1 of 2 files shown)


xsdg:
@caclark fyi

xsdg:
Merging this fix now to avoid blocking other pull requests (which would otherwise see a merge conflict if they're stacked atop this fix)

xsdg:
Cherry-picked as https://github.com/BestImageViewer/geeqie/commit/971d766c423375754e1e3abf744e0777d422426f

PR #2407, opened 2026-06-15
Updates `PanCacheData` to use `FileDataRef` to manage its contained `FileData *`

xsdg:
(Stacked on top of PR #2406)

Also creates a scripted test to trigger pan-view, along with performing some other minor code cleanups.

The scripted test doesn't do much, but should at least make sure that panview doesn't have any major runtime crashes.


xsdg:
@qarkai fyi

src/pan-view/pan-view.cc
```diff
@@ -110,9 +110,7 @@ constexpr GqColor PAN_POPUP_BORDER_COLOR{0, 0, 0, PAN_POPUP_ALPHA};
 
 void pan_cache_data_free(PanCacheData *pc)
 {
-	if (!pc) return;
-
-	file_data_unref(pc->fd);
+	// delete can handle nullptr, so no need to check.
```
qarkai, line 113: This comment is redundant. Otherwise we have to add it for every free/g_free/delete without nullptr check.

Consider removing this function. There is `delete_cb()` for using in `pan_cache_free()`.

src/pan-view/pan-view.cc
```diff
@@ -572,18 +569,18 @@ static gboolean pan_cache_step(PanWindow *pw)
...
 
 /* This sync date function is optimized for lists with a common sort */
 static void pan_cache_sync_date(const PanWindow *pw, GList *list)
 {
-	static const auto pan_cache_data_compare_fd = [](gconstpointer data, gconstpointer user_data)
+	static const auto pan_cache_data_compare_fd = [](gconstpointer data, gconstpointer user_data) -> int
 	{
 		auto *pc = static_cast<const PanCacheData *>(data);
-		return (pc->fd == user_data) ? 0 : 1;
+		return *(pc->fd_ref) == user_data;
```
qarkai, line 583: This changes logic. `true` converts to 1 which is 'not equal' for `GCompareFunc`.

src/pan-view/pan-view.cc
```diff
@@ -865,7 +862,7 @@ PanItemList pan_layout_intersect(PanWindow *pw, gint x, gint y, gint width, gint
 		GdkRectangle intersection;
 		if (!gdk_rectangle_intersect(&pg->rect, rect, &intersection)) return 1;
 
-		return gdk_rectangle_equal(rect, &intersection) ? 0 : 1;
+		return gdk_rectangle_equal(rect, &intersection);
```
qarkai, line 865: This changes logic. `FALSE` is 0.

src/pan-view/pan-view.cc
```diff
@@ -634,10 +631,10 @@ std::optional<GqSize> pan_cache_get_image_size(PanWindow *pw, const FileData *fd
 {
 	if (!fd) return {};
 
-	const auto pan_cache_data_cd_dimensions_compare_fd = [](gconstpointer data, gconstpointer user_data)
+	const auto pan_cache_data_cd_dimensions_compare_fd = [](gconstpointer data, gconstpointer user_data) -> int
 	{
 		auto *pc = static_cast<const PanCacheData *>(data);
-		return (pc->cd && pc->cd->dimensions && pc->fd == user_data) ? 0 : 1;
+		return pc->cd && pc->cd->dimensions && *(pc->fd_ref) == user_data;
```
qarkai, line 637: Also here.

qarkai (changes requested):
Replacement looks good.

Please revert comparison changes.

src/pan-view/pan-view.cc
xsdg: Thanks for catching these 🙏.  That said, it's interesting that I was playing around with the UI and didn't notice any breakages 🤔.  Will revert though

src/pan-view/pan-view.cc
xsdg: Replaced with `delete_cb`.  Thanks for pointing it out

src/pan-view/pan-view.cc
xsdg: Reverted

src/pan-view/pan-view.cc
xsdg: Reverted

PR #2410, opened 2026-06-15
Fixes `FD_VERBOSE_DEBUG` tests to switch on value, rather than presence.

xsdg:
Without this change, `FD_VERBOSE_DEBUG` behaves as if equivalent to `DEBUG_FILEDATA` (which is set in all debug builds)

Replaces PR #2404


caclark:
Implemented in commit c99fa87912d0bf0e4a1fcdb28e61d67ee669cf7d

PR #2411, opened 2026-06-15
Flattens and simplifies meson testing logic

xsdg:
A main goal here was to avoid the duplication of the `summary` lines that was happening with the nested `if` statements.

(Replaced PR #2405)


caclark:
Implemented in commit 276db1e2e509a5e50687a651576449179dc45a56

PR #2412, opened 2026-06-15
Refactors fd-refcount-test into a more generic scripted test library, and an implementation

xsdg:
(Stacked atop PR #2411)
The goal of this is to make it easy to turn any sequence of Geeqie `--action`s into a small UI test script.

(Replaces PR #2406)


caclark:
@xsdg 

Unfortunately I get an error when trying to apply this patch:

cclark@lr3:~/data/geeqie/geeqie$ git am ~/downloads/2412.patch
Applying: Flattens and simplifies meson testing logic
error: patch failed: meson.build:136
error: meson.build: patch does not apply
Patch failed at 0001 Flattens and simplifies meson testing logic

xsdg:
Yup, that always happens with these stacks (because github doesn't realize that some of the commits in each stacked PR were actually "owned by" the prior PR in the stack).  Will rebase and it should work.

xsdg:
@caclark  Rebased 👍

caclark:
Implemented in commit a36e1ee11b6f647f52b0968a60d981f74019c0b9 48fd81b95e1ada1c8cc5a431dca878fd34e17f45

PR #2415, opened 2026-06-16
Replace error-prone coding rule with C++ Core Guidelines rule

qarkai:


caclark:
Implemented in commit 86ccf29641ee3cceed36abb66934e1828b5f5657

qarkai:
Fix #1244.

PR #2416, opened 2026-06-17
Fix memory leak in editor_read_desktop_file()

qarkai:


caclark:
Implemented in commit 4f4a0bacd66893ba6374fafb7ad8edbf4d6eb0b5

PR #2417, opened 2026-06-17
Simplify filter_to_list() a bit

qarkai:
Use `g_strsplit()` instead of manual parsing.


caclark:
Implemented in commit 7b5f57efd6df6de3f5039d4071052fd0751e7a8e

PR #2418, opened 2026-06-18
Use get_editor_by_command() to reduce code duplication

qarkai:


caclark:
Implemented in commit ed51d94b29e2bd86a2524f3dbe929e2b3e324088

PR #2419, opened 2026-06-19
Reduce parameters number in editor functions

qarkai:


caclark:
Implemented in commit 6357a1df354be20c0eae4d6e57ec78112883da51

PR #2420, opened 2026-06-20
Use EditorData::vd as editor_verbose_io_cb() user data

qarkai:


caclark:
Implemented in commit e8b0873a22ccbf350f601de32ef73737bf6b2916

PR #2422, opened 2026-06-20
Fix double semicolon

qarkai:


caclark:
Implemented in commit baef8274e920b335e0afdcc85d572e21e843d89a

PR #2427, opened 2026-06-21
Return optional from generic_dialog_find_window()

qarkai:


caclark:
Implemented in commit af64a1ac56886612fdb834df3638c503c0340825

PR #2428, opened 2026-06-21
Use map for dialog_windows

qarkai:


src/ui-utildlg.cc
```diff
@@ -45,28 +44,19 @@
...
-	                       dialog_window_has_title_and_role);
+	if (title) key.first = title;
+	if (role) key.second = role;
 
-	return (it != dialog_windows.end()) ? *it : nullptr;
+	return key;
 }
 
-std::vector<DialogWindow *> dialog_windows;
+std::map<DialogWindowKey, GdkRectangle> dialog_windows;
```
xsdg, line 59: Does it matter that this won't maintain insert ordering like the `vector` had?  (I don't know)

qarkai: No, we don't rely on order anywhere.

caclark:
Implemented in commit 865e5df1da3b389778f4f24de215e5ee9497d6e6

Issue #2424, opened 2026-06-21
Startup crash: <dialogs> window list grows without bound


stapelberg:
**Setup (please complete the following information):**
 - Distribution: NixOS
 - Distribution release: 26.05
 - Geeqie version [geeqie --version]: Geeqie 2.5 GTK3

**Describe the bug**

I noticed geeqie got slow to start over the last few days.

When I wanted to start geeqie today, it just crashed:

```
% geeqie 2026-06-21-empty.jpg
Geeqie fatal error
Signal: Invalid memory reference
Code: Address not mapped
Address: 0x0000000000000018
geeqie(+0x104b89) [0x57ed5148eb89]
/nix/store/57iz36553175g3178pvxjij8z5rcsd4n-glibc-2.42-61/lib/libc.so.6(+0x42790) [0x7d0619642790]
geeqie(+0x15d6f4) [0x57ed514e76f4]
geeqie(_ZN8FileData27file_data_send_notificationEPS_10NotifyType+0x2f) [0x57ed514fbf7f]
geeqie(+0x16d91c) [0x57ed514f791c]
geeqie(_ZN8FileData13file_data_newEPKcP4statiP15FileDataContext+0x68) [0x57ed514f8a18]
geeqie(_ZN8FileData19file_data_new_groupEPKcP15FileDataContext+0x10f) [0x57ed514f973f]
geeqie(_Z15layout_set_pathP12LayoutWindowPKc+0x1b) [0x57ed5147a94b]
geeqie(_Z22layout_new_from_configPPKcS1_i+0xc4) [0x57ed5147f064]
geeqie(+0x129c17) [0x57ed514b3c17]
/nix/store/jlyahda14aya375lv7k9fsin2zk90nxz-glib-2.88.1/lib/libglib-2.0.so.0(+0x6a809) [0x7d061b308809]
/nix/store/jlyahda14aya375lv7k9fsin2zk90nxz-glib-2.88.1/lib/libglib-2.0.so.0(g_markup_parse_context_parse+0x1222) [0x7d061b309c82]
geeqie(_Z20load_config_from_bufPKcmi+0xe1) [0x57ed514b7281]
geeqie(_Z21load_config_from_filePKci+0x78) [0x57ed514b7398]
geeqie(_Z12load_optionsP11ConfOptions+0x41) [0x57ed51497211]
geeqie(main+0xd77) [0x57ed51425a77]
[...]

PR #2429, opened 2026-06-21
Removes constructor arguments that haven't been used in a long time

xsdg:
The intent here is to make `vd_new` easier to reason about (and particularly, the fact that `dir_fd` may remain uninitialized after `vd_new` returns.

The code that origially used this was refactored from `view_dir_{tree,list}.c` into `view_dir.c` in 5add0a6fd3900ca0c6ea0ed8989e74780f27a8c7

Pertains to #2424


caclark:
Implemented in commit e37008aad24c05ec091dba85a77c162e27f4d9ad

PR #2430, opened 2026-06-22
Refactors `FileDataRef` into its own header file

xsdg:
This is to make it easily usable from header files that forward-declare `FileData` instead of `#include`ing `filedata.h`.


caclark:
Implemented in commit 52397dd69d1c0451f041d08346386a46ee38e792

PR #2432, opened 2026-06-22
Switches `ViewData` implementations to use `FileDataRef` for `dir_fd` and temporaries

xsdg:
Also establishes and tests the `FileDataRef` equality behavior

This simplifies ref management, and fixes one of the missing `dir_fd` null checks in `view-data.cc` (which is now properly handled by the destructor)

Related to #2424

### DIFF
--- a/src/view-dir.cc
+++ b/src/view-dir.cc
@@ -173,23 +173,22 @@ static void vd_destroy_cb(GtkWidget *widget, gpointer data)
 		}
 
 	switch (vd->type)
-	{
-	case DIRVIEW_LIST: vdlist_destroy_cb(widget, data); break;
-	case DIRVIEW_TREE: vdtree_destroy_cb(widget, data); break;
-	}
+		{
+		case DIRVIEW_LIST: vdlist_destroy_cb(widget, data); break;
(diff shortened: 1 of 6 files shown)


src/view-dir.cc
```diff
@@ -173,15 +173,15 @@ static void vd_destroy_cb(GtkWidget *widget, gpointer data)
...
+		{
+		case DIRVIEW_LIST: vdlist_destroy_cb(widget, data); break;
+		case DIRVIEW_TREE: vdtree_destroy_cb(widget, data); break;
+		}
 
 	folder_icons_free(vd->pf);
 	file_data_list_free(vd->drop_list);
 
-	file_data_unref(vd->dir_fd);
+	vd->dir_fd.~FileDataRef();  // Manually destruct vd->dir_fd.
```
qarkai: I suggest to use `new`/`delete` for `ViewDir` instead of this.

xsdg: Good point.  Fixed.

qarkai (approved):

caclark:
Implemented in commits  c785d42fe8f9d052e89c9813bdacb7b45f8ee588 6150798ba3cd11e62754c63cc4387d7cd5304edf

PR #2436, opened 2026-06-24
Fix build

qarkai:


caclark:
Implemented in commit 63a57d546c769815dc2f847e1cd185eff23ee53c

PR #2461, opened 2026-07-01
Make spell build conditional

qarkai:
Use SPDX header in spell.cc/h


caclark:
Implemented in commit 6e2b1d258948523187d629c1008e87b6afc1df88

PR #2466, opened 2026-07-02
Adds a simple feature to allow capturing screenshots during the image tests

xsdg:
The goal is to make it easier to see what Geeqie is doing, without conflicting with other Geeqie instances that might be running on the same host.

Sample screenshot:
1783013292_IMG_0483_craw_CR3


caclark:
Implemented in commit 99886ceaf7efddf3c2f77f3308daf1927cd045ec

PR #2475, opened 2026-07-05
Use GtkAlertDialog for print error

qarkai:


caclark:
Implemented in commit ffd8380529426a77ee501be60c95dc9921c83393

PR #2476, opened 2026-07-05
Adds a utility to fire up a nested x-server that is independent of the current user's homedir.

xsdg:
Includes the ability to run a window manager and symlink contents into the isolated directory

Sample command:
`./tools/isolated_xserver.py --wm xfwm4 --link_path "imgs=${PWD}/build/test-images.p/images/" -- "${PWD}/build/src/geeqie"`

Screenshot (using Geeqie v2.7):
Screenshot_2026-07-05_21-46-32

### DIFF
--- a/tools/isolated_xserver.py
+++ b/tools/isolated_xserver.py
@@ -1,0 +1,179 @@
+#!/bin/python3
+# SPDX-License-Identifier: GPL-2.0-or-later
+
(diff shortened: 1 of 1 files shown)


caclark:
Implemented in commit d8c31f9b6c734f5a8e439b74594b496b1335e111

PR #2477, opened 2026-07-06
Fix adding markup to checkbox

qarkai:


caclark:
Implemented in commit b12fabd8691ad204b88859e4cc6a2e6129880249

PR #2480, opened 2026-07-08
meson: bump gtk4 dep to 4.20

tomaszg7:
It is required for `GDK_ACTION_NONE`, see https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gdk/gdkenums.h#L295

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -58,7 +58,7 @@ fs = import('fs')
 configuration_inc = include_directories('.')
 
 # Main library dependencies
-gtk_dep = dependency('gtk4', version : '>=4.18', required : true)
+gtk_dep = dependency('gtk4', version : '>=4.20', required : true)
 glib_dep = dependency('glib-2.0', version : '>=2.66', required: true)
 pango_dep = dependency('pango', version : '>=1.46', required: true)
 


gusnan:
Is there a solution that doesn't require bumping the gtk4 dependency? Debian stable only have GTK4 version 4.18.6. To me it seems like GDK_ACTION_NONE simply is a define to 0, but I don't know the implications of simply using 0 in this case.

tomaszg7:
No idea, I just noticed geeqie doesn't build for me anymore ;)

caclark:
@gusnan 

I have a question for you, because I think it may be easier for you to answer than for me.
Does the latest GTK4 branch compile under Debian Stable (apart from GDK_ACTION_NONE)?

gusnan:
@caclark 

It did just a few days ago, but if I now start to try to do a replace of the GDK_ACTION_NONE with 0 it seems pretty hard, some places don't take a 0 instead of a GDK_ACTION, so I have a pretty hard time making it comile. And the commits that added the GDK_ACTION_NONE seems pretty extensive too unfortunately.

I see that it's probably a huge work to get rid of them, so I understand if they will be kept. 

I do find it a pity that it wouldn't be buildable on older distros (like my Debian stable) though.

I'm starting to see the need for the GTK-ng project more and more - * sigh * ...

caclark:
@gusnan 

Please try inserting this into compat.h:


#ifndef GDK_ACTION_NONE
constexpr GdkDragAction GDK_ACTION_NONE = static_cast<GdkDragAction>(0);
#endif

gusnan:
@caclark That makes it build! Thanks!

gusnan:
@caclark With the change of the cast to `static_cast<GdkDragAction>(0);` , of course, sorry.

caclark:
There is a workaround in d8350c513baf0de14850be1a09ee80b9e03cd5d8

PR #2483, opened 2026-07-09
Remove bar callbacks wrappers

qarkai:


caclark:
Implemented in commit a93b674903fee6389f9c4190d876f4d01f86e258

PR #2493, opened 2026-07-13
Fix clipboard

qarkai:
Fix `GdkContentProvider` creation.
Fix memory leaks.


caclark:
Implemented in commit ca9008393fe432a8473dcba2b23a9f8d736126b4

PR #2496, opened 2026-07-14
Replace deprecated GtkComboBox in pan-view-filter

qarkai:


caclark:
Implemented in commit 64f0e3b06a2c97b444acf6ca714f6415d11a9b8c

PR #2500, opened 2026-07-18
Disable deprecated-declarations warnings

qarkai:
When there are too many warnings, it's just noise.
Remove `compat-deprecated.h`.
Fix other warnings.


caclark:
Implemented in commit 82c94e04d5f49511213e4a3ccdb9ad29fbab1376

PR #2503, opened 2026-07-19
Present the window on activation

City-busz:
This ensures that the window brings to the front if already open.


caclark:
Implemented in commit 34338fd477c51c474726a58a9e4aefefb7a89fd7

PR #2504, opened 2026-07-20
Wrap text on long line

City-busz:
This allows the preferences window to fit into smaller screens.


caclark:
Implemented in commit 63e13f095dc3ff454dabcc00cadd4469bb93e0e4

PR #2505, opened 2026-07-20
Use GqSize for ConfOptions::thumbnails::size

qarkai:


caclark:
Implemented in commit e464a9b90eb7ad0c58964a664a931a5a3bc2c3db

PR #2507, opened 2026-07-21
Use c_options as callbacks data

qarkai:


caclark:
Implemented in commit eb2d3f0166b37c139bd6053c0f728c6e60f480e4

PR #2519, opened 2026-07-30
Reduce usage of global c_options

qarkai:

PR #2520, opened 2026-07-31
Remove xvpics support

qarkai:


caclark:
Implemented in commit 8635ceded822621c2b1c9b78b3109fb65e18cdd7

xsdg:
Sorry, just for context, what's the motivation for this change?  (I don't doubt that it makes sense; I'm just curious what motivated it)

caclark:
I've got a real nose full of this endless GTK3/GTK4 migration, and have been using Codex. Whatever people's opinion of AI, without it the migration would never be completed.

I have been making continuous requests to Codex e.g. - "What next to do in the migration" - and going through step by step.

Codex was telling me xvpics  was history, and I went along with that.

It can be reversed if you would like. No problem. Maybe there are users who have xvpics thumbnails on data CDs.

[A few days ago I bought a used NAS unit - plugged some SATA drives into it - and the built-in OS wiped all the drives and replaced them with a RAID - without asking...
So, I've been looking through cardboard boxes in the garden shed for old 5.25 IDE drives and data CDs - not to mention having to find a SATA CD drive.
Maybe old stuff can be usefull]

xsdg:
No need to revert or anything; this sounds reasonable.  I was just asking because there wasn't much context to this PR; if the description had just said something like "xvpics is no longer generally used" I wouldn't have even asked.

qarkai:
See https://github.com/BestImageViewer/geeqie/pull/2519#issuecomment-5136056357

xsdg:
@qarkai Thanks for pointing that out; I had missed it.  I should probably update my github notifications config to make mentions more obvious.

PR #2523, opened 2026-08-03
Auto next folder

queenycrowley:
## Summary

Add an "Auto next folder" preference (Behavior tab, disabled by default). When enabled, reaching the last image in a folder automatically advances to the first image of the next sibling folder (alphabetical); reaching the first image goes back to the last image of the previous sibling folder.

Also adds explicit "Next folder" / "Previous folder" actions with Ctrl+Shift+Right / Ctrl+Shift+Left accelerators.

## Details
- New `ConfOptions.auto_next_folder` (default FALSE), persisted via rcfile
- Preferences checkbox in the Behavior tab
- `layout_get_next_sibling_dir()` / `layout_get_prev_sibling_dir()` helpers in layout-image.cc
- `FileData::FileList::read_list_lstat_all()` so hidden directories are considered as siblings
- Unit tests in tests/layout-image-navigation.cc (all pass)
- NEWS, README, and docbook help updated

Happy to adjust anything per maintainer feedback.

### DIFF
--- a/src/layout-image.cc
+++ b/src/layout-image.cc
@@ -1549,6 +1549,169 @@ std::optional<ColorManStatus> layout_image_color_profile_get_status(LayoutWindow
 	return image_color_profile_get_status(lw->image);
 }
 
+/**
+ * @brief Get the next sibling directory in the same parent directory
+ * @param lw Layout window
+ * @returns FileData for the next directory, or nullptr if none
+ *
+ * Finds the next directory alphabetically after the current directory
+ * in the same parent directory. Only returns directories that contain
+ * at least one image file.
+ */
+static FileData *layout_get_next_sibling_dir(LayoutWindow *lw)
+{
+	if (!lw || !lw->dir_fd || !lw->dir_fd->path) return nullptr;
+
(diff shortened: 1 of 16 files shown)


caclark:
@queenycrowley 

There is a Unit Test fail.

You can run tests locally by `./tools/test-all.sh`

caclark:
@queenycrowley 

Please do not update the NEWS file. That gets done only for a new release.

src/layout-image.cc
```diff
@@ -1549,6 +1549,169 @@ std::optional<ColorManStatus> layout_image_color_profile_get_status(LayoutWindow
...
+
+	return next_dir;
+}
+
+/**
+ * @brief Get the previous sibling directory in the same parent directory
+ * @param lw Layout window
+ * @returns FileData for the previous directory, or nullptr if none
+ */
+static FileData *layout_get_prev_sibling_dir(LayoutWindow *lw)
```
qarkai, line 1645: @queenycrowley 
Same as `layout_get_next_sibling_dir` but with `sort_settings.ascending = FALSE;`. I'm going to remove it.

qarkai: See #2577

PR #2524, opened 2026-08-04
Adds weston (Wayland) support to isolated_display helper

xsdg:
This makes it easier to see the Wayland experience from X11 (and hopefully, vice-versa)


caclark:
Implemented in commit 0e3d3c2065ad8c2f8c67d55fb14157f1b34e1208

PR #2527, opened 2026-08-04
Reduce code duplication in ViewFile filter callbacks

qarkai:


caclark:
Implemented in commit bcd53d9030c759408e81541946f1f7d535040cdd

PR #2530, opened 2026-08-06
Fix creating ViewDir

qarkai:


caclark:
Implemented in commit 9525100f1ddf45f4a0e84bf61163b185cc131f0b

PR #2544, opened 2026-08-11
Ignore -Wclobbered only for gcc

qarkai:
Clang doesn't support -Wclobbered.


caclark:
Implemented in commit 96a0195b28afd105e9a0d697a93d93d98d22a3ec

PR #2554, opened 2026-08-16
Use GDK_NO_MODIFIER_MASK

qarkai:


caclark:
Implemented in commit 92525ed6aa7215c065d41eb803ea789795118365

PR #2557, opened 2026-08-17
Use `CommitDate` in the version string

przepompownia:
Previously `AuthorDate` was used, which was confusing if the last merged commit was older than its parent.


caclark:
Implemented in commit 95cca2fcf8c0b72d5a70437713cbfa6442e1cb01

PR #2560, opened 2026-08-18
Remove get_alternative_button_order()

qarkai:
Remove now redundant `GenericDialog::cancel_button`.


caclark:
Implemented in commit cb11146879110d05de1a2552202414736b0a6580

PR #2563, opened 2026-08-20
Add ViewFileInfoIcon::tip_label to avoid static_cast-ing

qarkai:


caclark:
Implemented in commit 7e45a3e2e7d33cdb4737e0a1c0326b315416b48c

PR #2565, opened 2026-08-21
Simplify collection_table_selection_set() a bit

qarkai:


caclark:
Implemented in commit 27dfc248e76131d326ddc2291d96f14505eec32c

PR #2568, opened 2026-08-22
Use available enum values, reduce static_cast

qarkai:


caclark:
Implemented in commit 904998bbef41c27f1cbe975828b060e8e5610f40

PR #2573, opened 2026-08-24
Fix show_editor_list_window()

qarkai:


caclark:
Implemented in commit 5768ca211042443ca8ebeae9bed27f092ced8942

PR #2575, opened 2026-08-25
Update German translation

eishorn:
Tried to keep German translation up to date. New to this thing here. Used Poedit and replaced the whole file, so diff seems kinda messed up.

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -11,379 +11,426 @@ msgid ""
 msgstr ""
 "Project-Id-Version: geeqie-1.3\n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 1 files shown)


caclark:
@eishorn 

Sorry. 

Your PR made me look at the .po file generation. It was messed up. I have made a couple of commits to fix that.
Your Pull Request now fails on the latest de.po.

Please revise and re-request.

caclark:
Fixed in commit 6cb9d5773300406c69207928252d4e03f6e20849

@eishorn 

I used Codex to fix your patch. You might check a couple of entries to see if it has done a reasonable job.

eishorn:
Okay thank you, will do. Could be a few days, though.

PR #2580, opened 2026-08-30
New Update to de.po

eishorn:
Reworked German translation again, also incorporating the fixes of @jahangmar.


caclark:
Implemented in commit 59273e122fe98d141e0e68a12bb56852fa3c4c8d

PR #2589, opened 2026-09-02
Make more use of save_config_to_file() options parameter

qarkai:


caclark:
Implemented in commit bdadb3168ced4b69a07ca380916ee2da1ee89b8a

PR #2595, opened 2026-09-08
Use gdk_pixbuf_read_pixels() for read-only pixbufs

qarkai:
Avoid implicit copy in `gdk_pixbuf_get_pixels()`.


caclark:
Implemented in commit 18bb433dfea35e714e4ee72547a373f17e0e3b8e

PR #2598, opened 2026-09-10
Optimize file_data_verify_ci() a bit

qarkai:
No need to iterate whole list.


caclark:
Implemented in commit e04b1cec104053cec50b75bdfa2e61f6faf9a67f

PR #2599, opened 2026-09-10
Use GHashTable for FileData::cached_metadata

qarkai:


caclark:
Implemented in commit d59378999a5cc42044254dc248530545d2523aca

PR #2600, opened 2026-09-13
Add RcString to encapsulate rcfile writing

qarkai:


qarkai:
Probably we should use dedicated XML library to avoid manual indentation.

caclark:
> Probably we should use dedicated XML library to avoid manual indentation.

I have given up thinking. I let Codex do that....

---

I wouldn't add libxml2 just for this

Looking at the actual code changes my recommendation somewhat. I'd keep GMarkup.

The reading side is already using an appropriate library and is quite lightweight. The custom GQParserData dispatch mechanism is doing useful application-specific work rather than reimplementing XML.

The writing side is hand-written, but the XML structure is simple and controlled entirely by Geeqie. Introducing libxml2 merely to replace:

WRITE_STRING("<foo ");
WRITE_CHAR(...);
WRITE_STRING(">");
...
WRITE_STRING("</foo>");

with xmlTextWriter*() calls adds a dependency without eliminating the large amount of configuration-specific code.

There is potentially a much more worthwhile refactoring, though: the repeated WRITE_* and READ_* lists could be driven by a table describing each option—name, type, member/accessor, constraints, perhaps compatibility/deprecation information. That could eliminate a substantial amount of the paired read/write boilerplate and, importantly, make it harder for the reader and writer to get out of sync. That would buy you much more than replacing GMarkup with libxml2.

qarkai:
libxml2 is not supported. I was thinking of PugiXML, TinyXML2 or something. Unfortunately GLib has only XML parser, not creator. So, yes, it means additional dependency.

It's not direct WRITE_SOMETHING() -> xmlTextWriterSomething() replacement. We build XML document and dump it to file. The main advantage is we don't care of closing tags and indentation. Probably we can implement this without XML library, just using RcString. I don't know how yet. Maybe add wrappers or something.

caclark:
Is this Pull Request still valid irrespective of future actions concerning xml?

qarkai:
> Is this Pull Request still valid irrespective of future actions concerning xml?

Yes.

caclark:
Implemented in commit d2aacdd9c2ed293fcab54afc5575d1bc8996d01f

xsdg:
For what it's worth, I would be generally in favor of using an XML library to produce XML.  I feel like getting rid of custom parsers and serializers is a positive, and I feel like an XML serializer will likely be useful for us, in light of GTK4 moving in the direction of using XML more extensively than GTK3 did.

PR #2610, opened 2026-09-17
Add default PaneData::pane_event implementation

qarkai:
Note: `bar_event()` is not used at the moment.


caclark:
Implemented in commit d924e8f13801665029624c86bfde48d2a6de71a2

PR #2612, opened 2026-09-18
Return optional from metadata_read_GPS_coord()

qarkai:
Avoid magic number for invalid value.


caclark:
Implemented in commit 430e33fa8d9fb2a7e4eb5f8090c496e542a7d373

PR #2614, opened 2026-09-20
Fix memory leak in bar_pane_keywords_destroy()

qarkai:


caclark:
Implemented in commit 1af26dbeb9e82539436b9b1d6c08d07c23736daf

Issue #2617, opened 2026-09-21
Missing default window layout file notification


przepompownia:
**Setup (please complete the following information):**
 - Distribution: Debian
 - Distribution release: Sid
 - Geeqie version: 3.2+git20260921-b0cd5990 GTK4

**Describe the bug**
Geeqie warns (once) about missing `default_window_layout.xml`. I haven't bisected what commit causes this warning and when it occurs in existing config directories, but it is simple to reproduce on empty environment like in #2616. On the next start the notification is not shown although there is still no such file. 

**To reproduce**
Prepare empty `.xdg/config` directory and run `XDG_CONFIG_HOME=$PWD/.xdg/config geeqie`.

**Expected behavior**
From the user side - no warning, especially on an empty configuration.

przepompownia:
There is yet one path of this issue: on next start this notification appears when trying to create new collection, and the new Geeqie window appears (to drag and drop into new collection as I suspect).

przepompownia:
Thanks! When checking the change I got similar notification: `geeqierc.xml - Failed to open file “.../.xdg/config/geeqie/geeqierc.xml”: No such file or directory`.

PR #2622, opened 2026-09-22
Check if config file exists before loading

przepompownia:
I guess we need a similar check like in https://github.com/BestImageViewer/geeqie/commit/0154a2c44efaf6f85c41efcf6f432a1893c3fae5

See https://github.com/BestImageViewer/geeqie/issues/2617#issuecomment-5777827447


caclark:
Implemented in commit 584b25e9b03b9ee08a0d9fe91a2e863373227ccf

